PR #34, opened 2022-06-21
Update one function used in MVPC according to the slides (page 14)

TURuibo:
1. Update the code line 232 - 246: naming, line 241 - 245: not empty according to the slides.
2. Add a test file for the function.


TURuibo:
About the test, I first generate a dataset with missing values in the MAR way. And then apply the function to detect the parents of the missingness indicators. The following is the notebook visualization and the results.
Screenshot 2022-06-21 at 11 32 47
Screenshot 2022-06-21 at 11 32 59

Issue #38, opened 2022-06-24
How to handle named variables


AndrewC19:
Hi,

I have been working with the causal-learn package for a couple of weeks now and I was wondering if there is a plan to add support for named variables? 

As I understand it, at the moment data is generally assumed to take the form of a 2d np.array where each column contains a value for a particular variable `X1, X2, ..., XN`. I can quite easily take a csv file and convert it to this format but, in doing so, I lose the names of my columns. This relies on me having to keep track of which variable (`X1, X2, ..., XN`) corresponds to which named column of my csv data, which makes tasks such as visualising the output quite difficult.

I have been using the `to_pyplot` method to draw the discovered DAG and passing it the labels. However, this relies on the order being preserved (simply replacing X1 with the first label, X2 with the second, and so on). This has tripped me a couple of times when comparing two graphs produced from similar data where the variables do not align.

I am not sure what the solution is here, but it would be really useful if it were possible to attach actual variable names to the graph rather than doing this manually after computation. One possibility would be to support pandas dataframes for reading in data.

kunwuz:
Hi Andrew, thanks for your interest, and we are very grateful for your question. Just a quick confirmation, regarding the `to_pyplot` method you mentioned, did you mean `GraphUtils.to_pydot(cg.G, labels=[“A”, “B”, “C”])`?

AndrewC19:
Hi, yes that's what I mean.

kunwuz:
Thanks for the clarification! we will get back to you with potential solutions quickly :)

AndrewC19:
Great, thanks. Let me know if any further information would be helpful.

AndrewC19:
Hi again, I have encountered another similar issue that may influence your solution.

If my csv of data contains columns of different type (say three continuous variables for age, height, and weight, and categorical variables for gender and blood group), loading them into a numpy array forces you to encode them as the same type. This means that if my categorical variables are encoded as ints they will all be cast to floats, or if they are encoded as strings, all of my continuous data will be cast to strings.

This is problematic for dealing with categorical variables automatically (I would like to apply a different conditional independence test depending on this). Ideally, I would treat all floats as continuous and then use a heuristic to determine if my integers should be treated as continuous or categories. However, all of my integers are being converted to floats so I have no reliable way of differentiating them.

I'm not sure if there is a way to encode the type of the variable in causal-learn already. If so, please could you show me how?

kunwuz:
[...]

PR #42, opened 2022-06-28
support user input variables name to PC

tofuwen:
# Description
This PR adds node_names parameters to PC and change the underlying functions accordingly. Now users can specify node_names when running PC.

This PR is intended to fix the following issue that raised by user: https://github.com/cmu-phil/causal-learn/issues/38

# Test Plan

Run the following code to make sure we get expected results.
```
data_path = "tests/TestData/data_linear_10.txt"
data = np.loadtxt(data_path, skiprows=1)  # Import the file at data_path as data
names = [f"tofu{i+1}" for i in range(data.shape[1])]
cg = pc(data, 0.05, fisherz, False, 0, -1, node_names=names)  # Run PC and obtain the estimated graph (CausalGraph object)
cg.draw_pydot_graph()
```

Screen Shot 2022-06-28 at 2 54 13 PM

### DIFF
--- a/causallearn/graph/GraphClass.py
+++ b/causallearn/graph/GraphClass.py
@@ -21,8 +21,11 @@ from causallearn.utils.PCUtils.Helper import list_union, powerset
 
 
 class CausalGraph:
-    def __init__(self, no_of_var: int):
-        node_names: List[str] = [("X%d" % (i + 1)) for i in range(no_of_var)]
+    def __init__(self, no_of_var: int, node_names: List[str] | None = None):
+        if node_names is None:
+            node_names = [("X%d" % (i + 1)) for i in range(no_of_var)]
+        assert len(node_names) == no_of_var, "number of node_names must match number of variables"
+        assert len(node_names) == len(set(node_names)), "node_names must be unique"
         nodes: List[Node] = []
         for name in node_names:
             node = GraphNode(name)
--- a/causallearn/search/ConstraintBased/PC.py
+++ b/causallearn/search/ConstraintBased/PC.py
@@ -16,29 +16,48 @@ from causallearn.utils.PCUtils.BackgroundKnowledgeOrientUtils import \
     orient_by_background_knowledge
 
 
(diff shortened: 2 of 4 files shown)


MarkDana (commented):
Hi Yewen! This is looking good. I ran the test example and it is correct. But I have some comments:

Now `node_names` parameter is added to the functions `pc` => `mvpc_alg`, `pc_alg` => `skeleton_discovery`, in order to pass it to the `CausalGraph` instance created in `skeleton_discovery`. However, this seems to be over coupling - `node_names` is redundant for functions other than `pc`.

Is it possible that we add a function in `CausalGraph` class that changes its nodes names, and here we only have the `node_names` parameter in `pc`. After the `CausalGraph` instance (with nodes names `"X%d"`) is returned by `mvpc_alg` or `pc_alg`, we then change its names before it is returned to the user. This seems can be adapted to FCI, CD-NOD, and GES more easily, I guess?

Thanks Yewen for your work :))

tofuwen:
Hi @MarkDana,

Thanks for your awesome review! :)

1. IMHO, the CausalGraph constructor needs to take node_names as its input parameters --- it makes more sense to initial the nodes with their correct names. We shouldn't initial with wrong names and change it later.
2. mvpc_alg and pc_alg are only used in pc, so I think adding a parameter int those functions is fine? Regarding skeleton_discovery, it is used in more places --- but any function that return a CausalGraph should return a correct graph (i.e. logically it should contain a CausalGraph that shouldn't need more post-process like changing node_names)
3. change node_names in graph may have side effects --- if I remember correctly, sometimes some code will use names as node identifiers. Without good tests, I am afraid doing so will have some troubles. (And I still think initially with wrong names and change it later doesn't seem to logically consistent.) :)

Hope this makes sense and we can discuss more! :)

ztz1989:
Hi @tofuwen and @MarkDana,

Could you do the same for GES? I have a ground truth graph built with txt2generalgraph; the node names in the file do not follow the "X%d" format. So it is hard to use the estimation functions to compute the accuracy and SHD.

Thanks!

PR #44, opened 2022-06-30
Added some unit tests for PC

MarkDana:
### Updated files:
`tests/TestPC.py`, and some .txt files (data, graph, benchmark results, etc.) to load in `tests/TestData`.

### Usage example:
```python
python -m unittest TestPC.TestPC.test_pc_load_linear_10_with_fisher_z
```

### What these unit tests can do:
+ Contains tests for PC over loaded dataset (linear Gaussian, discrete, and linear Gaussian with missing values), and PC over data simulated from specific graphs.
+ For developers, make sure their modified code is logically consistent (return the same result as benchmark) with the benchmark (as of commit 94d1536).
+ For users, know how to run PC, and have a quick glance at the PC algorithms performance (e.g. SHD to truth graph).

### TODO:
+ Design more comprehensive fuzzing analysis, including
  - For code, traverse over every possible execution flow of PC.
  - For data, design dataset of corner cases (e.g., with super big cardinalities/sample size).
+ Incorporate `TestMVPC*.py`.
+ Incorporate PC with background knowledge.
+ Need to explain why (and how) users should choose parameters (of `stable`, `uc_rule`, and `uc_priority`) as in this test file.
+ Reorganize files in `tests/` folder.


MarkDana:
Hi @tofuwen Thanks so much for your comments. Learned a lot :)

Regarding your questions:

> I am a little worried about the correctness of current PC algorithm.

Me too. And thus it's very likely that the benchmark result files require frequent changes in the near future (as stated in the comment block [here](https://github.com/MarkDana/causal-learn/blob/d1c0ccf2bb1b08a4764f57ecdf42429984f6addb/tests/TestPC.py#L14)). I am still thinking of a convenient way to update those benchmark files and the corresponding MD5 verifications at every usage (e.g. [here](https://github.com/MarkDana/causal-learn/blob/d1c0ccf2bb1b08a4764f57ecdf42429984f6addb/tests/TestPC.py#L85)).

> Is it possible to run Tetrad on some dataset as well and compare it.

Great suggestion! I'll do it soon. Currently there already seems to be some results returned by Tetrad java code (e.g. [tetrad_graph_discrete_10.txt](https://github.com/cmu-phil/causal-learn/blob/main/tests/TestData/tetrad_graph_discrete_10.txt)) in `TestData`. But I don't know what are the respective configuration that generate them. @jdramsey @kunwuz Maybe you happen to know? Thanks!

> Is it possible to add more simple test cases

Cool! I'll do so. And we may also need more complex test cases - it may just happen to return the correct graph on simple cases.

> For files under true_dag_graph, are those the "true graph"? (we have true graph information? Where did we get that?)

I found them at e.g. [graph.10.txt](https://github.com/cmu-phil/causal-learn/blob/main/tests/TestData/graph.10.txt). But similar as above, I don't know the respective codes to generate datasets. @jdramsey @kunwuz How about we also include codes to generate e.g. data_linear_10.txt from graph.10.txt?

> How long it takes to finish each test roughly?

Except for `test_pc_load_bnlearn_discrete_datasets`, each should cost less than 20s. `test_pc_load_bnlearn_discrete_datasets` should cost around one minute. **TODO** I'll update the exact time later.

> Could you include a test plan section which contains which tests you run and the results? (The result screenshot) 

It's a nice suggestion for general unit tests, but for the tests included in this PR, if I understand correctly, it seems that we only need to pass each tests? - there is no need to verify or compare results by human, since they are done by assertions.

Thanks so much for your greeeatt work👍!

tofuwen:
1. Yeah, you are perfect right on "we may also need more complex test cases". It's always better to have more tests! :) But I think that may need additional work --- and adding some simple tests to cover more code paths (e.g. different independence test; different variable types) is already great, and can already make sure at least at the code level, our code is mostly likely correct (and can detect stupid bugs already). :)
2. Regarding test plan section, we still need to include it in every PR, saying what did you do to test this PR. For this PR, in your test plan section, it can be as simple as writing all the test commands you ran (i.e. python -m unittest TestPC.TestPC.test_pc_load_linear_10_with_fisher_z), and includes a screenshot of your terminal saying the test passed (the screenshot is optional though --- you can just say in your test plan all the tests passed)
3. I also commented on fixing random seed in your PR, maybe you miss it? :) @MarkDana

PR #9, opened 2021-12-23
Debug: 1. cache refresh at every init; 2. compute cardinality only once

MarkDana:
### Bugs solved:
1. About cache refresh:
 + Manually refresh cache at every init. Added one refresh line at `SkeletonDiscovery.py:L145` (for PC) and `FCI.py:L615` (for FCI).
 + Deprecate `hash(data.tobytes())` - it's slow. @chenweiDelight also mentions a faster `hash(str(data))` - but let's just init refresh and use no hash.

2. About passing cardinality chisq or gsq:
 + Added `cardinalities` and `is_discrete` at `Fas.py:L10`, and calculate cardinalities only once at `SkeletonDiscovery.py:L165` (for PC) and `FCI.py:L625` (for FCI). No need to `np.max()` every time.

### TODO: still debugging:
Now PC results with/without fas (change `PC.py:L71`) respectively:

| data (#nodes/#edges)<br>time (sec) | without fas time | with fas time | SHD(without fas, with fas) |
|:----------------------------------:|:----------------:|:-------------:|:--------------------------:|
|             cancer 5/4             |       0.009      |     0.009     |              0             |
|           earthquake 5/4           |       0.012      |      0.01     |              0             |
|             survey 6/6             |       0.014      |     0.015     |              0             |
|              asia 8/8              |       0.024      |     0.025     |              0             |
|             sachs 11/17            |       0.142      |     0.148     |              0             |
|             child 20/25            |       0.618      |     0.699     |              4             |
|           insurance 27/52          |       1.45       |     1.823     |              3             |
|             water 32/66            |       0.321      |     0.365     |              0             |
|             alarm 37/46            |       0.873      |     1.015     |              1             |
|            barley 48/84            |       3.466      |      4.95     |              0             |
|          hailfinder 56/66          |       0.938      |     1.588     |              0             |
|            hepar2 70/123           |       9.482      |     11.949    |              3             |
|           win95pts 76/112          |       3.381      |     4.504     |              0             |
|            andes 223/338           |      26.741      |     45.722    |              0             |

So two problems:
1. Still small SHD difference at datasets e.g. `child`.
2. Still time difference, e.g. `andes` 26s vs 45s - though before solving cardinalities issue, it's ~300s.
3. See profiling stat: callings count on chisq is different on with/without fas.

PR #46, opened 2022-07-02
Rewrite CITests as a class && re-use covariance matrix for fisherz

MarkDana:
### Updated files:

+ `cit.py`: main change: integrate CITests functions to a class
+ `PC.py`, `FCI.py`, `CDNOD.py`, `Fas.py`, `SkeletonDiscovery.py`: adjust accordingly for the new CIT class.
+ `GraphClass.py`: remove unnecessary attributes (e.g., data, cache) in the `CausalGraph` class.
+ `TestPC.py`: there is randomness in `mvpc`, and thus we need two lines of test: w/ or wo/ randomness.
+ `TestFCI.py`, `TestMVPC_mv_fisherz_test.py`: minor issues (e.g., file directory).

### Why this update is needed:
+ **1. Speedup.** In current `pc` and `fisherz`, the correlation matrix over a same dataset `data` is re-computed at every call on `fisherz`, which wastes lots of time. Thanks @Yujia for raising this speedup!
+ However, if we simply add a `if indep_test == fisherz` line and pass the pre-computed `corr_mat` to `fisherz` at every possible usage, code will be super lengthy and ugly: `indep_test` is used everywhere in constraint-based methods, and parameters passing will be complicated.
+ An example of above: for speedup, we added CITest cache and cardinalities of gsq/chisq tests [before](https://github.com/cmu-phil/causal-learn/pull/9). However, to achieve this, these parameters are passed all across modules, e.g. [here](https://github.com/cmu-phil/causal-learn/blob/104b5dd3c9bf8991b22335ed65df2522b545e524/causallearn/search/ConstraintBased/FCI.py#L44), [here](https://github.com/cmu-phil/causal-learn/blob/104b5dd3c9bf8991b22335ed65df2522b545e524/causallearn/utils/PCUtils/SkeletonDiscovery.py#L62), and [here](https://github.com/cmu-phil/causal-learn/blob/104b5dd3c9bf8991b22335ed65df2522b545e524/causallearn/utils/Fas.py#L59).
+ **2. To make further changes on CITests easier**. Now all attributes of a specific CItests is stored in its CIT object, including `data`, `cache`, and: `corr_mat` for `fisherz`, `cardinalities` for `gsq/chisq`, user-specified parameters for `kci`, etc.

### How to use the CIT class
+ For users, any algorithms can be run end-to-end in a same way as before, e.g.,

```python
from causallearn.search.ConstraintBased.PC import pc
from causallearn.utils.cit import fisherz

cg = pc(data, 0.05, fisherz)
```
though they may notice that now `causallearn.utils.cit.fisherz` is a string `"fisherz"`, instead of the function before.

+ For developers, now we should write code as:

```python

fisherz_obj = CIT(data, "fisherz") # construct a CIT instance with data and method name
pValue = fisherz_obj(X, Y, S) # a simple call is ok. no need to consider cache/corr_mat etc. by yourself.
```

while before we write code as:

```python
from causallearn.utils.cit import fisherz

# some complex if..then.. considerations before calculation, e.g., corr_mat, cache.
pValue = fisherz(data, X, Y, S)
```

### Test plan:
To ensure the new CIT class is correct and does not change the logic of the original code (as of commit 94d1536):

```bash
python -m unittest TestPC    # should pass
python -m unittest TestFCI    # should pass
python -m unittest TestCDNOD    # should pass
python -m unittest TestMVPC    # should pass
python -m unittest TestMVPC_mv_fisherz_test.py    # should pass
python -m unittest TestMVPC_mv_fisherz.py    # should pass except for test_pc_with_mv_fisherz_MCAR_data_assertion, which cannot pass in the original code.
```

### Speedup gain:
E.g., for `TestPC.TestPC.test_pc_with_fisher_z`, it takes `24.812s` in a run of the original code. Now it takes `3.748s`.

### Todo:
+ `cit.py: L113`: `np.corr` numerical issues with `nan` values.
+ To affect as less code as possible, now we deleted unnecessary attributes e.g., `data`, `cache` in the `CausalGraph` class. While we still have `test` attributes (a `CIT` object), to adapt to `cg.ci_test` methods in the original codes. But, for simplicity, should `test` be part of a graph class, eventually?

### DIFF
--- a/causallearn/graph/GraphClass.py
+++ b/causallearn/graph/GraphClass.py
@@ -18,6 +18,7 @@ from causallearn.graph.GraphNode import GraphNode
 from causallearn.graph.Node import Node
 from causallearn.utils.GraphUtils import GraphUtils
 from causallearn.utils.PCUtils.Helper import list_union, powerset
+from causallearn.utils.cit import CIT
 
 
 class CausalGraph:
@@ -34,10 +35,7 @@ class CausalGraph:
         for i in range(no_of_var):
             for j in range(i + 1, no_of_var):
                 self.G.add_edge(Edge(nodes[i], nodes[j], Endpoint.TAIL, Endpoint.TAIL))
-
-        self.data = None  # store the data
-        self.test = None  # store the name of the conditional independence test
-        self.corr_mat = None  # store the correlation matrix of the data
+        self.test: CIT | None = None
         self.sepset = np.empty((no_of_var, no_of_var), object)  # store the collection of sepsets
         self.definite_UC = []  # store the list of definite unshielded colliders
         self.definite_non_UC = []  # store the list of definite unshielded non-colliders
@@ -47,35 +45,17 @@ class CausalGraph:
         self.nx_skel = nx.Graph()  # store the undirected graph
         self.labels = {}
         self.prt_m = {}  # store the parents of missingness indicators
-        self.mvpc = False
-        self.cardinalities = None  # only works when self.data is discrete, i.e. self.test is chisq or gsq
-        self.is_discrete = False
-        self.citest_cache = dict()
-        self.data_hash_key = None
-        self.ci_test_hash_key = None
 
-    def set_ind_test(self, indep_test, mvpc=False):
+
+    def set_ind_test(self, indep_test):
         """Set the conditional independence test that will be used"""
-        # assert name_of_test in ["Fisher_Z", "Chi_sq", "G_sq"]
-        self.mvpc = mvpc
         self.test = indep_test
-        self.ci_test_hash_key = hash(indep_test)
 
     def ci_test(self, i: int, j: int, S) -> float:
         """Define the conditional independence test"""
         # assert i != j and not i in S and not j in S
-        if self.mvpc:
-            return self.test(self.data, self.nx_skel, self.prt_m, i, j, S)
-
-        i, j = (i, j) if (i < j) else (j, i)
-        ijS_key = (i, j, frozenset(S), self.data_hash_key, self.ci_test_hash_key)
-        if ijS_key in self.citest_cache:
-            return self.citest_cache[ijS_key]
-        # if discrete, assert self.test is chisq or gsq, pass into cardinalities
-        pValue = self.test(self.data, i, j, S, self.cardinalities) if self.is_discrete \
-            else self.test(self.data, i, j, S)
-        self.citest_cache[ijS_key] = pValue
-        return pValue
+        if self.test.method == 'mc_fisherz': return self.test(i, j, S, self.nx_skel, self.prt_m)
+        return self.test(i, j, S)
 
     def neighbors(self, i: int):
         """Find the neighbors of node i in adjmat"""
--- a/causallearn/search/ConstraintBased/CDNOD.py
+++ b/causallearn/search/ConstraintBased/CDNOD.py
@@ -10,9 +10,10 @@ from causallearn.utils.PCUtils import SkeletonDiscovery, UCSepset, Meek
 from causallearn.utils.PCUtils.BackgroundKnowledge import BackgroundKnowledge
 from causallearn.utils.PCUtils.BackgroundKnowledgeOrientUtils import orient_by_background_knowledge
 from causallearn.utils.cit import *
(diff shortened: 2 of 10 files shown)


tofuwen:
I have some general design comments about this PR, how about we find some time to discuss?

1. It would be better to split the diff into smaller chunks (actually each commit should be a single PR to make the review easier --- this PR have about 1300 significant lines, and we should target at <200 lines in each PR). In industry, we can use things called "stacked diff", i.e. it allows you to send out new PR based on your previous PR (instead of current code trunk), so splitting PR wouldn't block you. If you split PRs, when you old PR get accepted, you can push them first.

2. For huge PR like this, it would be good to go through the design first then implement. Otherwise, we may still need some huge changes. :)

causallearn/search/ConstraintBased/PC.py
```diff
@@ -34,8 +35,7 @@ def pc(
         warnings.warn("The number of features is much larger than the sample size!")
 
     if mvpc:  # missing value PC
-        if indep_test == fisherz:
-            indep_test = mv_fisherz
+        if indep_test == fisherz: indep_test = mv_fisherz
```
tofuwen, line 38: hmmm, should have two lines here

tests/TestPC.py
```diff
@@ -4,12 +4,15 @@
 import unittest
 import hashlib
 import numpy as np
+np.random.seed(42)
```
tofuwen, line 7: how about we have two tests here?

Also please do not fix random seed for whole file here.

How about we have one test with seed fixed, one test without seed fixed? And only fix seed in the fix-seed function?

tofuwen (commented):
I didn't spend time reviewing cit.py --- I guess it's mainly copy and paste and we have cit tests already?

I think the PR is generally good to push --- only some nits comments.

Thanks the great work! It not only fix the old issues, but also significantly increased the code quality. :)

causallearn/graph/GraphClass.py
```diff
@@ -34,10 +35,7 @@ def __init__(self, no_of_var: int, node_names: List[str] | None = None):
         for i in range(no_of_var):
             for j in range(i + 1, no_of_var):
                 self.G.add_edge(Edge(nodes[i], nodes[j], Endpoint.TAIL, Endpoint.TAIL))
-
-        self.data = None  # store the data
-        self.test = None  # store the name of the conditional independence test
-        self.corr_mat = None  # store the correlation matrix of the data
```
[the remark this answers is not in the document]
MarkDana: I verified this by PyCharm -> Find Usages and there is no other usages in the project.

Is this a reliable way to find usages or do you have any other recommendations?

causallearn/search/ConstraintBased/CDNOD.py
```diff
@@ -10,9 +10,10 @@
 from causallearn.utils.PCUtils.BackgroundKnowledge import BackgroundKnowledge
 from causallearn.utils.PCUtils.BackgroundKnowledgeOrientUtils import orient_by_background_knowledge
 from causallearn.utils.cit import *
+from causallearn.search.ConstraintBased.PC import get_parent_missingness_pairs, skeleton_correction
 
 
-def cdnod(data: ndarray, c_indx: ndarray, alpha: float = 0.05, indep_test=fisherz, stable: bool = True,
+def cdnod(data: ndarray, c_indx: ndarray, alpha: float = 0.05, indep_test: str = fisherz, stable: bool = True,
```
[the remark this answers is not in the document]
MarkDana: Cool! Fixed.

causallearn/search/ConstraintBased/PC.py
MarkDana: √

tests/TestPC.py
MarkDana: √

causallearn/graph/GraphClass.py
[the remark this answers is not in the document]
tofuwen: I don't know. Maybe try it using other variables to see whether it did a great job? Normally these kind of things should rely on simple tests (if we have 100% test coverage hhh) && smart IDE.

tofuwen:
Great, after fixing all the nits above, I think this PR is ready to go!

PR #49, opened 2022-07-04
KCI mild speedup

MarkDana:
### Updated functions:
+ `causallearn/utils/KCI/KCI.py`:
   - `KCI_UInd.get_kappa`: reduce O(n^3) to O(n^2) based on the equation `np.trace(K.dot(K)) == np.sum(K * K.T)`.
   - L429 `elif self.kernelY == 'Polynomial'`: fixed a naming bug in original code. Should be `kernelZ`.
   - `KCI_CInd.KCI_V_statistic`: L479, reduced one repeated calculation based on the fact that `Kzx` and `Kzy` are usually same.
 + `causallearn/utils/KCI/Kernel.py`:
   - `Kernel.center_kernel_matrix`: reduce O(n^3) to O(n^2) by plugging `H = eye(n) - 1.0 / n` into `H.dot(K.dot(H))` and preventing the matrix multiplication.
 + `causallearn/search/FCMBased/lingam/hsic.py`: similar trick as `Kernel.center_kernel_matrix`.
 + `causallearn/utils/cit.py`: some trivial issue about specifying KCI kwargs.

### Speedup gain:
+ `KCI_UInd` (unconditional test): huge speedup (order of magnitude).
+ `KCI_CInd` (conditional test): reduces around 30% of time.
+ Overall, since conditional tests is always the bottleneck in an algorithm, this speedup is mild.
+ And thus we'll need more efficient way to calculate inverse and eigens of big matrix.

### Test plan:
+ To ensure that this update is logically consistent with the original code (under any possible parameters):
```python
np.random.seed(42)
from causallearn_fa79007.utils.cit import CIT as CIT_new
from causallearn_97e03ff.utils.cit import CIT as CIT_old

data = np.random.uniform(-1, 1, (100, 4))
for kernelname in ['Gaussian', 'Polynomial', 'Linear']:
    for est_width in ['empirical', 'median', 'manual']:
        for kwidth in [0.05, 0.1, 0.2]:
            for use_gp in [True, False]:
                for approx in [True, False]:
                    for polyd in [1, 2]:
                        kci_new = CIT_new(data, 'kci', kernelX=kernelname, kernelY=kernelname, kernelZ=kernelname,
                                              est_width=est_width, use_gp=use_gp, approx=approx, polyd=polyd,
                                              kwidthx=kwidth, kwidthy=kwidth, kwidthz=kwidth)
                        kci_old = CIT_old(data, 'kci', kernelX=kernelname, kernelY=kernelname, kernelZ=kernelname,
                                              est_width=est_width, use_gp=use_gp, approx=approx, polyd=polyd,
                                              kwidthx=kwidth, kwidthy=kwidth, kwidthz=kwidth)

                        # since there is randomness in null_sample_spectral, we need to fix the same seed for old and new run
                        np.random.seed(42); new_pval = kci_new(0, 1)
                        np.random.seed(42); old_pval = kci_old(0, 1)
                        assert np.isclose(old_pval, new_pval), "KCI_UIND is inconsistent after update."

                        np.random.seed(42); new_pval = kci_new(0, 1, (2, 3))
                        np.random.seed(42); old_pval = kci_old(0, 1, (2, 3))
                        assert np.isclose(old_pval, new_pval), "KCI_CIND is inconsistent after update."
```

+ Test speedup for `KCI_UInd`:
```python
for samplesize in range(1000, 20001, 1000):
    data = np.random.uniform(-1, 1, (samplesize, 2))
    kci_new = CIT_new(data, 'kci')
    kci_old = CIT_old(data, 'kci')
    tic = time.time(); kci_new(0, 1); tac = time.time(); time_new = tac - tic
    tic = time.time(); kci_old(0, 1); tac = time.time(); time_old = tac - tic
    print(f'{samplesize}:   {time_new} {time_old}')
```

<details>
<summary>Click to expand result</summary>

1000:   0.021124839782714844 0.051096200942993164
2000:   0.09719705581665039 0.2551460266113281
3000:   0.2547168731689453 0.7614071369171143
4000:   0.41847896575927734 1.7734618186950684
5000:   0.6685688495635986 3.47568416595459
6000:   0.9402382373809814 5.513365030288696
7000:   1.2306361198425293 9.101417064666748
8000:   1.7108919620513916 13.641188144683838
9000:   2.544398069381714 19.632148027420044
10000:   2.4818150997161865 25.683223009109497
11000:   2.8643081188201904 34.334508180618286
12000:   3.705733060836792 42.83501696586609
13000:   4.3541929721832275 56.730401039123535
14000:   4.609248161315918 68.94092202186584
15000:   5.336583852767944 83.44993829727173
16000:   6.201767921447754 103.21840572357178
17000:   7.315479040145874 128.12122511863708
18000:   8.262160062789917 153.92690014839172
19000:   8.924943208694458 182.9679470062256
20000:   9.806303977966309 210.9531922340393
</details>

time_uind

+ Test speedup for `KCI_CInd`:
```python
for samplesize in range(500, 10000, 250):
    data = np.random.uniform(-1, 1, (samplesize, 4))
    kci_new = CIT_new(data, 'kci')
    kci_old = CIT_old(data, 'kci')
    tic = time.time(); kci_new(0, 1, (2, 3)); tac = time.time(); time_new = tac - tic
    tic = time.time(); kci_old(0, 1, (2, 3)); tac = time.time(); time_old = tac - tic
    print(f'{samplesize}:   {time_new} {time_old}')
```

<details>
<summary>Click to expand result</summary>

500:   0.07363319396972656 0.10682296752929688
750:   0.23387503623962402 0.3343160152435303
1000:   0.4689028263092041 0.6680917739868164
1250:   1.0056912899017334 1.4335689544677734
1500:   1.8545763492584229 2.555225133895874
1750:   2.874561071395874 4.148720979690552
2000:   4.18735408782959 5.866267681121826
2250:   6.662433862686157 9.67330002784729
2500:   9.315114974975586 12.643889904022217
2750:   12.94197392463684 17.26453423500061
3000:   15.682774782180786 22.286062002182007
3250:   19.692520141601562 27.745235919952393
3500:   24.944950103759766 35.524142265319824
3750:   31.416036128997803 44.23967480659485
4000:   37.706125020980835 52.7396559715271
4250:   47.966763973236084 65.09382581710815
4500:   54.48931813240051 75.27616381645203
4750:   62.57163095474243 87.28439497947693
5000:   73.18962788581848 103.77545189857483
5250:   83.41915202140808 117.27962899208069
5500:   94.05676174163818 132.2629098892212
5750:   108.3281478881836 151.59633588790894
6000:   126.17060780525208 180.97449898719788
6250:   136.30685591697693 191.5569679737091
6500:   151.8123619556427 212.14035725593567
6750:   170.19479298591614 239.7194480895996
7000:   199.10978388786316 270.4898717403412
7250:   206.63027906417847 290.0352849960327
7500:   226.2251410484314 315.22972893714905
7750:   248.7306571006775 345.4090187549591
8000:   290.0406291484833 413.12121295928955
8250:   313.7044348716736 450.7824430465698
8500:   332.1730182170868 466.674519777298
8750:   352.3237581253052 486.6480839252472
9000:   374.6881170272827 519.5066709518433
9250:   412.67669320106506 571.1528761386871
9500:   436.2383370399475 612.9915819168091
9750:   474.7570939064026 669.191967010498
</details>

time_cind

### DIFF
--- a/causallearn/utils/KCI/KCI.py
+++ b/causallearn/utils/KCI/KCI.py
@@ -26,7 +26,6 @@ class KCI_UInd(object):
     [2] A. Gretton, K. Fukumizu, C.-H. Teo, L. Song, B. Schölkopf, and A. Smola, "A kernel
        Statistical test of independence." In NIPS 21, 2007.
     """
-
     def __init__(self, kernelX='Gaussian', kernelY='Gaussian', null_ss=1000, approx=True, est_width='empirical',
                  polyd=2, kwidthx=None, kwidthy=None):
         """
@@ -73,10 +72,15 @@ class KCI_UInd(object):
         _________
         pvalue: p value (scalar)
         test_stat: test statistic (scalar)
+
+        [Notes for speedup optimization]
+            Kx, Ky are both symmetric with diagonals equal to 1 (no matter what the kernel is)
+            Kxc, Kyc are both symmetric
         """
 
         Kx, Ky = self.kernel_matrix(data_x, data_y)
         test_stat, Kxc, Kyc = self.HSIC_V_statistic(Kx, Ky)
+
         if self.approx:
             k_appr, theta_appr = self.get_kappa(Kxc, Kyc)
             pvalue = 1 - stats.gamma.cdf(test_stat, k_appr, 0, theta_appr)
@@ -210,10 +214,18 @@ class KCI_UInd(object):
         _________
         k_appr, theta_appr: approximated parameters of the gamma distribution
 
+        [Updated @Haoyue 06/24/2022]
+        equivalent to:
+            var_appr = 2 * np.trace(Kx.dot(Kx)) * np.trace(Ky.dot(Ky)) / T / T
+        based on the fact that:
+            np.trace(K.dot(K)) == np.sum(K * K.T), where here K is symmetric
+        we can save time on the dot product by only considering the diagonal entries of K.dot(K)
+        time complexity is reduced from O(n^3) (matrix dot) to O(n^2) (traverse each element),
+        where n is usually big (sample size).
         """
         T = Kx.shape[0]
-        mean_appr = np.trace(Kx) * np.trace(Ky) / T
-        var_appr = 2 * np.trace(Kx.dot(Kx)) * np.trace(Ky.dot(Ky)) / T / T
+        mean_appr = np.diag(Kx).sum() * np.diag(Ky).sum() / T # same as np.trace(Kx) * np.trace(Ky) / T. a bit faster
+        var_appr = 2 * np.sum(Kx ** 2) * np.sum(Ky ** 2) / T / T # same as np.sum(Kx * Kx.T) ..., here Kx is symmetric
         k_appr = mean_appr ** 2 / var_appr
         theta_appr = var_appr / mean_appr
         return k_appr, theta_appr
@@ -426,11 +438,11 @@ class KCI_CInd(object):
                 # construct Gaussian kernels according to learned hyperparameters
                 Kzy = gpy.kernel_.k1(data_z, data_z)
                 self.epsilon_y = np.exp(gpy.kernel_.theta[-1])
-        elif self.kernelY == 'Polynomial':
+        elif self.kernelZ == 'Polynomial':
             kernelZ = PolynomialKernel(self.polyd)
             Kzx = kernelZ.kernel(data_z)
             Kzy = Kzx
-        elif self.kernelY == 'Linear':
+        elif self.kernelZ == 'Linear':
             kernelZ = LinearKernel()
             Kzx = kernelZ.kernel(data_z)
             Kzy = Kzx
@@ -453,9 +465,21 @@ class KCI_CInd(object):
         Vstat: KCI v statistics
         KxR: centralized kernel matrix for data_x (nxn)
         KyR: centralized kernel matrix for data_y (nxn)
+
+        [Updated @Haoyue 06/24/2022]
+        1. Kx, Ky, Kzx, Kzy are all symmetric matrices.
+            - Kx, Ky are with diagonal elements of 1 (because of exp(-0.5 * sq_dists * self.width)).
+            - If (self.kernelZ == 'Gaussian' and self.use_gp), then Kzx (Kzy) has all the same diagonal elements (not necessarily 1).
+              Otherwise Kzx, Kzy are with diagonal elements of 1.
+        2. If not (self.kernelZ == 'Gaussian' and self.use_gp): assert (Kzx == Kzy).all()
+           With this we could save one repeated calculation of pinv(Kzy+\epsilonI), which consumes most time.
         """
-        KxR = Kernel.center_kernel_matrix_regression(Kx, Kzx, self.epsilon_x)
-        KyR = Kernel.center_kernel_matrix_regression(Ky, Kzy, self.epsilon_y)
+        KxR, Rzx = Kernel.center_kernel_matrix_regression(Kx, Kzx, self.epsilon_x)
+        if self.epsilon_x != self.epsilon_y or (self.kernelZ == 'Gaussian' and self.use_gp):
+            KyR, _ = Kernel.center_kernel_matrix_regression(Ky, Kzy, self.epsilon_y)
+        else:
+            # assert np.all(Kzx == Kzy), 'Kzx and Kzy are the same'
+            KyR = Rzx.dot(Ky.dot(Rzx))
         Vstat = np.sum(KxR * KyR)
         return Vstat, KxR, KyR
 
--- a/causallearn/utils/KCI/KCI.py
+++ b/causallearn/utils/KCI/KCI.py
@@ -224,7 +224,7 @@ class KCI_UInd(object):
         where n is usually big (sample size).
         """
         T = Kx.shape[0]
-        mean_appr = np.diag(Kx).sum() * np.diag(Ky).sum() / T # same as np.trace(Kx) * np.trace(Ky) / T. a bit faster
+        mean_appr = np.trace(Kx) * np.trace(Ky) / T
         var_appr = 2 * np.sum(Kx ** 2) * np.sum(Ky ** 2) / T / T # same as np.sum(Kx * Kx.T) ..., here Kx is symmetric
         k_appr = mean_appr ** 2 / var_appr
         theta_appr = var_appr / mean_appr
--- a/causallearn/utils/KCI/Kernel.py
+++ b/causallearn/utils/KCI/Kernel.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 from abc import abstractmethod
 
+import numpy as np
 from numpy import eye, shape, ndarray
 from numpy.linalg import pinv
 
@@ -25,10 +26,19 @@ class Kernel(object):
     def center_kernel_matrix(K: ndarray):
         """
         Centers the kernel matrix via a centering matrix H=I-1/n and returns HKH
+        [Updated @Haoyue 06/24/2022]
+        equivalent to:
+            H = eye(n) - 1.0 / n
+            return H.dot(K.dot(H))
+        since n is always big, we can save time on the dot product by plugging H into dot and expand as sum.
+        time complexity is reduced from O(n^3) (matrix dot) to O(n^2) (traverse each element).
+        Also, consider the fact that here K (both Kx and Ky) are symmetric matrices, so K_colsums == K_rowsums
         """
+        # assert np.all(K == K.T), 'K should be symmetric'
         n = shape(K)[0]
-        H = eye(n) - 1.0 / n
-        return H.dot(K.dot(H))
+        K_colsums = K.sum(axis=0)
+        K_allsum = K_colsums.sum()
+        return K - (K_colsums[None, :] + K_colsums[:, None]) / n + (K_allsum / n ** 2)
 
     def center_kernel_matrix_regression(K: ndarray, Kz: ndarray, epsilon: float):
         """
@@ -36,4 +46,4 @@ class Kernel(object):
         """
         n = shape(K)[0]
         Rz = epsilon * pinv(Kz + epsilon * eye(n))
-        return Rz.dot(K.dot(Rz))
+        return Rz.dot(K.dot(Rz)), Rz
--- a/causallearn/utils/cit.py
+++ b/causallearn/utils/cit.py
@@ -31,10 +31,13 @@ class CIT(object):
 
         if method == 'kci':
             # parse kwargs contained in the KCI method
-            kci_kwargs = {k: v for k, v in kwargs.items() if k in
-                          ['kernelX', 'kernelY', 'null_ss', 'approx', 'est_width', 'polyd', 'kwidthx', 'kwidthy']}
-            self.kci_ui = KCI_UInd(**kci_kwargs)
-            self.kci_ci = KCI_CInd(**kci_kwargs)
+            kci_ui_kwargs = {k: v for k, v in kwargs.items() if k in
+                             ['kernelX', 'kernelY', 'null_ss', 'approx', 'est_width', 'polyd', 'kwidthx', 'kwidthy']}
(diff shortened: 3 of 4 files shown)


jdramsey:
Nice.

kunwuz:
Wow, that's really incredible! Thanks so much, Haoyue!!!

A dumb question: any idea why `np.sum(K * K.T)` is much faster for calculating the trace of the square?

tofuwen:
Wow, this is great, @MarkDana! Thanks for identifying such a non-trivial optimization!! :)

tofuwen:
@kunwuz to answer your question, matrix multiplication is in general O(n^3) (but there is some optimization usually, but I am not expert in this area), and K * K.T is only O(n^2), so the optimization above reduced from O(n^3) to O(n^2)

kunwuz:
> @kunwuz to answer your question, matrix multiplication is in general O(n^3) (but there is some optimization usually, but I am not expert in this area), and K * K.T is only O(n^2), so the optimization above reduced from O(n^3) to O(n^2)

Thanks! Yeah, I always thought that NumPy did a perfect job in optimizing matrix multiplication so I didn't expect the empirical speed-up will be such huge. Nice to know that trick~ :)

kunwuz:
> @kunwuz That's very different. Even NumPy did a perfect job at optimizing matrix multiplication, there is no way to reduce matrix multiplication to O(n^2). Here the problem is that we only need the diagonal entries of K.dot(K), but we calculated the whole K.dot(K), which wastes lots of time. Probably the optimization cannot be at NumPy level --- the optimization needs to at the "compile" level, i.e. compiler needs to know the program only need diagonal entries, so underlying machine level code it compiled to does NOT do what you told it to do, i.e. it didn't even calculate K.dot(K) for you. Of course, this is just a rough idea but not 100% correct --- I don't know the mechanmism of Python much, so the language I use is mostly for C.

Yes, you are totally right. The hypothesized best possible value for the exponent \omega is 2, so we cannot exactly achieve it. If I remember correctly, there exists optimization that achieved \omega < 2.4, and numpy is one of the SOTA methods, which goes beyond compiler level (BLAS), in optimizing matrix multiplication. So I'm a little bit surprised the EMPIRICAL improvement could be such huge, which is really helpful for me.

Also, for this specific optimization of the trace of the square, we don't need to calculate K.dot(K) at all. The only thing we need to take care of is how to efficiently get the diagonal entries. Thanks so much for your brilliant answer! :)

MarkDana:
Oh @kunwuz I just saw your message! Yes BLAS is a very carefully finetuned library. For the matrix multiplication complexity, [wikipedia](https://en.wikipedia.org/wiki/Matrix_multiplication_algorithm) says that "As of December 2020, the matrix multiplication algorithm with best asymptotic complexity runs in O(n2.3728596) time" - cool, let's see how this \omega will be pushing forward.

And thanks so much @tofuwen for your nice explanation - that's exactly what I identified. `np.trace(K.dot(K))` only needs the diagonal elements of `K.dot(K)`, so we can save time on the non-diagonal elements calculation.

And @tofuwen's comment is very insightful: "Probably the optimization cannot be at NumPy level --- the optimization needs to at the "compile" level" - to let the complier determine the computational graph, not exactly as the code we write. Actually now numpy is already smart enough to detect some of these cases (though not at complier level; need your specification in codes). For example:

We know that the matrix multiplication `M @ M.T` returns a symmetric matrix. Can numpy detect this and save time by calculating only the triangular half? The answer is yes:

```python
In [1]: import numpy as np
In [2]: M = np.random.rand(5000, 5000)
In [3]: MT_copy = M.T.copy()

In [4]: timeit M @ M.T
281 ms ± 3.69 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [5]: timeit M @ MT_copy
496 ms ± 4.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

But, for some of the calculations that I thought would be optimized - they are actually not. For example, there is no special treatment to `zeros` or `eye` in matrix multiplication:

```python
In [6]: zeros = np.zeros_like(M)
In [7]: timeit M @ zeros
497 ms ± 3.49 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [8]: eye = np.eye(5000)
In [9]: timeit M @ eye
507 ms ± 11.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

I don't know the specific reason for this lol :)

kunwuz:
@MarkDana Aha, thanks for this information! I will keep this in mind when I implement stuff in the future :-)

causallearn/utils/KCI/KCI.py
```diff
@@ -210,10 +214,18 @@ def get_kappa(self, Kx, Ky):
...
+        based on the fact that:
+            np.trace(K.dot(K)) == np.sum(K * K.T), where here K is symmetric
+        we can save time on the dot product by only considering the diagonal entries of K.dot(K)
+        time complexity is reduced from O(n^3) (matrix dot) to O(n^2) (traverse each element),
+        where n is usually big (sample size).
         """
         T = Kx.shape[0]
-        mean_appr = np.trace(Kx) * np.trace(Ky) / T
-        var_appr = 2 * np.trace(Kx.dot(Kx)) * np.trace(Ky.dot(Ky)) / T / T
+        mean_appr = np.diag(Kx).sum() * np.diag(Ky).sum() / T # same as np.trace(Kx) * np.trace(Ky) / T. a bit faster
```
tofuwen, line 227: curious why this will be faster than directly calling np.trace()?

causallearn/utils/KCI/KCI.py
```diff
@@ -426,11 +438,11 @@ def kernel_matrix(self, data_x, data_y, data_z):
                 # construct Gaussian kernels according to learned hyperparameters
                 Kzy = gpy.kernel_.k1(data_z, data_z)
                 self.epsilon_y = np.exp(gpy.kernel_.theta[-1])
-        elif self.kernelY == 'Polynomial':
+        elif self.kernelZ == 'Polynomial':
```
tofuwen, line 441: this seems an important bug to me --- the old code doesn't really work I guess?

causallearn/utils/KCI/KCI.py
```diff
@@ -453,9 +465,21 @@ def KCI_V_statistic(self, Kx, Ky, Kzx, Kzy):
...
+        2. If not (self.kernelZ == 'Gaussian' and self.use_gp): assert (Kzx == Kzy).all()
+           With this we could save one repeated calculation of pinv(Kzy+\epsilonI), which consumes most time.
         """
-        KxR = Kernel.center_kernel_matrix_regression(Kx, Kzx, self.epsilon_x)
-        KyR = Kernel.center_kernel_matrix_regression(Ky, Kzy, self.epsilon_y)
+        KxR, Rzx = Kernel.center_kernel_matrix_regression(Kx, Kzx, self.epsilon_x)
+        if self.epsilon_x != self.epsilon_y or (self.kernelZ == 'Gaussian' and self.use_gp):
+            KyR, _ = Kernel.center_kernel_matrix_regression(Ky, Kzy, self.epsilon_y)
+        else:
+            # assert np.all(Kzx == Kzy), 'Kzx and Kzy are the same'
```
tofuwen, line 481: why this assertion is commented? We should have assertions whenever possible

causallearn/utils/KCI/Kernel.py
```diff
@@ -25,15 +26,24 @@ def centering_matrix(n: int):
...
         Centers the kernel matrix via a centering matrix H=I-1/n and returns HKH
+        [Updated @Haoyue 06/24/2022]
+        equivalent to:
+            H = eye(n) - 1.0 / n
+            return H.dot(K.dot(H))
+        since n is always big, we can save time on the dot product by plugging H into dot and expand as sum.
+        time complexity is reduced from O(n^3) (matrix dot) to O(n^2) (traverse each element).
+        Also, consider the fact that here K (both Kx and Ky) are symmetric matrices, so K_colsums == K_rowsums
         """
+        # assert np.all(K == K.T), 'K should be symmetric'
```
tofuwen, line 37: why not keep this assertion?

causallearn/utils/cit.py
```diff
@@ -31,10 +31,13 @@ def __init__(self, data, method='fisherz', **kwargs):
...
-                          ['kernelX', 'kernelY', 'null_ss', 'approx', 'est_width', 'polyd', 'kwidthx', 'kwidthy']}
-            self.kci_ui = KCI_UInd(**kci_kwargs)
-            self.kci_ci = KCI_CInd(**kci_kwargs)
+            kci_ui_kwargs = {k: v for k, v in kwargs.items() if k in
+                             ['kernelX', 'kernelY', 'null_ss', 'approx', 'est_width', 'polyd', 'kwidthx', 'kwidthy']}
+            kci_ci_kwargs = {k: v for k, v in kwargs.items() if k in
+                             ['kernelX', 'kernelY', 'kernelZ', 'null_ss', 'approx', 'use_gp', 'est_width', 'polyd',
+                              'kwidthx', 'kwidthy', 'kwidthz']}
+            self.kci_ui = KCI_UInd(**kci_ui_kwargs)
+            self.kci_ci = KCI_CInd(**kci_ci_kwargs)
```
tofuwen, line 40: hmmm, seems a bug to me previously.

Are the old runnable when calling KCI_CInd?

tofuwen:
Thanks for the great work, I think the code is ready to push after addressing all the nits.

causallearn/utils/KCI/KCI.py
MarkDana: Oh my fault. Last time I checked (but maybe with something wrong) and thought that `np.diag(Kx).sum()` is faster by sacrificing the space of new array `np.diag(Kx)` created.

I just double checked. The two's speed performance is almost same. I'll change back to `np.trace`. Thanks!

causallearn/utils/KCI/KCI.py
MarkDana: Yes it's a typo bug. `KernelY` should be `KernelZ`.

However, in practical usage, people usually choose the same kernel for both X, Y, and for Z. And thus this bug happens to be avoided so far.

causallearn/utils/KCI/KCI.py
MarkDana: Just for speed. `np.all(Kzx == Kzy)` takes O(n^2) time to traverse each elements (just the same as the key function part e.g. `center_kernel_matrix` after optimization).

I had assertions when writing and testing. But I tried to remove them for deployment. What do you think?

causallearn/utils/cit.py
MarkDana: My fault! The old code is runnable when calling KCI_CInd, but not KCI_UInd, since there are arguments e.g., `kernelZ`, `use_gp` specifically for KCI_CInd and there is no **kwargs in KCI_UInd.

To prevent from this again, I'll add a unit test using KCI in `TestPC.py` soon.

aoqiz (approved):
Looks good to me. Thanks for this great work.

causallearn/utils/KCI/KCI.py
tofuwen: Cool, good to know! 

If we have tests earlier, we would detect this bug. ;)

@aoqiz A good case why good tests are needed. :)

causallearn/utils/KCI/KCI.py
tofuwen: cool, sounds good! Will any user input break this assertion? If so, maybe we can still add this assertion to make sure user gives the correct input.

causallearn/utils/cit.py
tofuwen: awesome, thanks!!

causallearn/utils/KCI/KCI.py
aoqiz: Yes, thanks.

causallearn/utils/KCI/KCI.py
MarkDana: To the user end, any data input or parameters choice will not break this assertion, i think.

PR #57, opened 2022-07-13
Added some unit tests for Astar and DP

ignavierng:
### Updates
- Added test cases for Astar and DP
- Added simulated data and ground truth for the test cases
- Removed previous test cases for Astar and DP

### Test Plan
`python -m unittest tests.TestAstar # should pass`
`python -m unittest tests.TestDP # should pass`

TestAstar
TestDP


tests/TestAstar.py
```diff
@@ -1,42 +1,60 @@
...
-            dag_est, search_stats = bic_exact_search(X, search_method='astar')
-            print(dag_est)
-            print(search_stats)
+
+class TestAstar(unittest.TestCase):
+    # Load data and run Astar with default parameters.
+    def test_astar_simulate_linear_gaussian_with_local_score_BIC(self):
+        print('Now start test_astar_simulate_linear_gaussian_with_local_score_BIC ...')
+        truth_CPDAG_matrix = np.loadtxt("./TestData/test_exact_search_simulated_linear_gaussian_CPDAG.txt")
+        data = np.loadtxt("./TestData/test_exact_search_simulated_linear_gaussian_data.txt")
```
tofuwen, line 43: How do we generate this data? Do you mind adding the data generating script in this file as well (but comment out the code) --- this can ensure we know how the data is generated later.

tofuwen (commented):
awesome work!

Only some very nits need to be addressed. :)

tofuwen:
In your screenshot, I saw 0,1,2,3 are printed. Do we need to print them? Maybe we can remove it?

tests/TestAstar.py
ignavierng: Thanks for the suggestion. I have added and commented the code to generate the data and ground truth.

tofuwen:
Yeah, please change the dag2cpdag. We don't the need the extra print in test, which is confusing.

And in any industry level code, there shouldn't be print statement haha

ignavierng:
> Yeah, please change the dag2cpdag. We don't the need the extra print in test, which is confusing.
> 
> And in any industry level code, there shouldn't be print statement haha

I notice that the reason is not the `dag2cpdag` function. The issue was that the tests I wrote were importing causallearn from my pip installed packages instead of the local code, which was not desired. I have fixed it. Thanks!

tofuwen (approved):
looks good! Thanks for the great work!

cc @kunwuz to merge this

PR #59, opened 2022-07-15
Refactored unit tests for Astar and DP

ignavierng:
### Updates
- Refactored unit tests for Astar and DP into a single function for exact search

### Test Plan
`python -m unittest tests.TestExactSearch # should pass`

TestExactSearch


tofuwen (approved):
Thanks, great work! 
Btw, you will also change other tests to remove the sys.path.append(), right?

cc @kunwuz to merge this PR.

ignavierng:
Yes, I will change the other tests to update the data path and remove `sys.path.append()`, so that in the future we should perhaps run `python -m unittest tests.FileName`, rather than `cd`-ing into `tests` folder and running `python -m unittest FileName`, the latter which imports causallearn (if exists) from pip installed packages instead of local one.

PR #60, opened 2022-07-16
add local_score_BIC_from_cov

wean2016:
### Description section
1. add local_score_BIC_from_cov
2. set the default bic score function in GES.py with local_score_BIC_from_cov

### Test plan section
python -m unittest TestGES.py # should pass
image

### DIFF
--- a/causallearn/score/LocalScoreFunction.py
+++ b/causallearn/score/LocalScoreFunction.py
@@ -41,6 +41,40 @@ def local_score_BIC(Data: ndarray, i: int, PAi: List[int], parameters=None) -> f
     return n * H + np.log(n) * len(PAi) * lambda_value
 
 
+def local_score_BIC_from_cov(
+    Data: Tuple[ndarray, int], i: int, PAi: List[int], parameters=None
+) -> float:
+    """
+    Calculate the *negative* local score with BIC for the linear Gaussian continue data case
+    from the covariance matrix
+    Parameters
+    ----------
+    Data: covariance matrix, number of instances
+    i: current index
+    PAi: parent indexes
+    parameters: lambda_value, the penalty discount of bic
+    Returns
+    -------
+    score: proportional to -2x BIC
+    """
+
+    cov, n = Data
+
+    if parameters is None:
+        lambda_value = 1
+    else:
+        lambda_value = parameters["lambda_value"]
+
+    if len(PAi) == 0:
+        return n * np.log(cov[i, i])
+
+    yX = np.mat(cov[np.ix_([i], PAi)])
+    XX = np.mat(cov[np.ix_(PAi, PAi)])
+    H = np.log(cov[i, i] - yX * np.linalg.inv(XX) * yX.T)
+
+    return n * H + np.log(n) * len(PAi) * lambda_value
+
+
 def local_score_BDeu(Data: ndarray, i: int, PAi: List[int], parameters=None) -> float:
     """
     Calculate the *negative* local score with BDeu for the discrete case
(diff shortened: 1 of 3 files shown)


tofuwen:
Could you update your description to be more informative? It can provide reviewers more context.

In the description, it's better to indicate why we need this change, what is this change supposed to do, the benefits of this new functionality. :)

tofuwen:
Hmm, we need a better design for this PR and in general LocalScoreClass. We should arrange a meeting to talk about the design. The current way is not extensible --- in the near future, it would be super difficult to maintain.

Some general suggestions when writing code:
1. you should try to avoid copy and paste code (or never do it). I noticed that, the code in your new function is basically the same as previous function --- you should try to reuse code.
2. Think about how to re-use other part of the code in the codebase --- I am not 100% sure, but is it possible to combine LocalScoreClass with CIT class in some way? It seems to me they are doing similar thing: using data to calculate something, and cache / cov matrix should be re-used. cc @MarkDana So let's have a joint meeting to discuss.

kunwuz:
> Hmm, we need a better design for this PR and in general LocalScoreClass. We should arrange a meeting to talk about the design. The current way is not extensible --- in the near future, it would be super difficult to maintain.
> 
> Some general suggestions when writing code:
> 
>     1. you should try to avoid copy and paste code (or never do it). I noticed that, the code in your new function is basically the same as previous function --- you should try to reuse code.
> 
>     2. Think about how to re-use other part of the code in the codebase --- I am not 100% sure, but is it possible to combine LocalScoreClass with CIT class in some way? It seems to me they are doing similar thing: using data to calculate something, and cache / cov matrix should be re-used. cc @MarkDana So let's have a joint meeting to discuss.

Thanks for the great suggestions! Yes, I totally agree with you that the score function, as long as many other modules, should be refactored as previously mentioned.

At the same time, it might also be necessary to first make some critical/quick changes that benefit the algorithms a lot. For example, in this PR, we could directly use the covariance matrix as the input to avoid repeat computation, which should be included ASAP to improve the performance. But of course, we should never forget to refactor afterward for maintainability and further changes :)

tofuwen:
yeah, @kunwuz is right.

@wean2016, when you have time, we can book a meeting and I can share with you some general coding principles which may help you write code in the future. :) We can use this PR as an example.

wean2016:
> yeah, @kunwuz is right.
> 
> @wean2016, when you have time, we can book a meeting and I can share with you some general coding principles which may help you write code in the future. :) We can use this PR as an example.

Thank you very much for your suggestions, I think you are very right. We should write better-designed code. Also, I've been catching up on my paper lately and I'm a bit busy. Could we wait for the AAAI submission to finish before we meet about this topic?

wean2016:
> @wean2016 Sure, this sounds reasonable.
> 
> Could you add a todo (and assign the todo to you) in the code (format: # @todo (your name)) to remind us to refactor after you have more time? Then I think we can push this PR first to at least unblock our users. :)

Yes, I agree with your opinions. I add some # todo in the score function file. Then, this change may remind us in the future.

MarkDana:
> Hmm, we need a better design for this PR and in general LocalScoreClass. We should arrange a meeting to talk about the design. The current way is not extensible --- in the near future, it would be super difficult to maintain.
> 
> Some general suggestions when writing code:
> 
> 1. you should try to avoid copy and paste code (or never do it). I noticed that, the code in your new function is basically the same as previous function --- you should try to reuse code.
> 2. Think about how to re-use other part of the code in the codebase --- I am not 100% sure, but is it possible to combine LocalScoreClass with CIT class in some way? It seems to me they are doing similar thing: using data to calculate something, and cache / cov matrix should be re-used. cc @MarkDana So let's have a joint meeting to discuss.

Sure! Happy to discuss :)

tofuwen:
Sure, looks good. I think we can merge this PR for now. @kunwuz

PR #61, opened 2022-07-18
Added some unit tests for GIN

zhi-yi-huang:
## Description
The GIN algorithm returns the graph and the causal order. In this algorithm, the causal order and the causal graph correspond to each other, so only the causal order in the result is asserted.

## Updates

- Fixed the independence test problem in GIN
- Added assertion to the previous test cases
- Removed the plot function in previous test cases for GIN
- Added tests to test GIN algorithm using the hsic independence test.
- Updated docstring
- Removed default parameter of indep_test
- Refactored the test code

## Test Plan

```python
python -m unittest tests.TestGIN # should pass
```
image

### DIFF
--- a/causallearn/search/HiddenCausal/GIN/GIN.py
+++ b/causallearn/search/HiddenCausal/GIN/GIN.py
@@ -27,8 +27,8 @@ def GIN(data, indep_test_method='kci', alpha=0.05):
     ----------
     data : numpy ndarray
            data set
-    indep_test : str, default='kci'
-        the function of the independence test being used
+    indep_test_method : str, default='kci'
+        the name of the independence test being used
     alpha : float, default=0.05
         desired significance level of independence tests (p_value) in (0,1)
     Returns
@@ -47,7 +47,7 @@ def GIN(data, indep_test_method='kci', alpha=0.05):
     if indep_test_method not in ['kci', 'hsic']:
         raise NotImplementedError((f"Independent test method {indep_test_method} is not implemented."))
 
-    def indep_test(x, y, method=indep_test_method):
+    def indep_test(x, y, method):
         if method == 'kci':
             return kci.compute_pvalue(x, y)[0]
         elif method == 'hsic':
@@ -65,7 +65,7 @@ def GIN(data, indep_test_method='kci', alpha=0.05):
             e = cal_e_with_gin(data, cov, list(cluster), list(remain_var_set))
             pvals = []
             for z in range(len(remain_var_set)):
-                pvals.append(indep_test(data[:, [z]], e[:, None]))
+                pvals.append(indep_test(data[:, [z]], e[:, None], method=indep_test_method))
             fisher_pval = fisher_test(pvals)
             if fisher_pval >= alpha:
                 tmp_clusters_list.append(cluster)
@@ -96,7 +96,7 @@ def GIN(data, indep_test_method='kci', alpha=0.05):
                 e = cal_e_with_gin(data, cov, X + cluster_i1 + cluster_j1, Z + cluster_i2)
                 pvals = []
                 for z in range(len(Z + cluster_i2)):
-                    pvals.append(indep_test(data[:, [z]], e[:, None]))
+                    pvals.append(indep_test(data[:, [z]], e[:, None], method=indep_test_method))
                 fisher_pval = fisher_test(pvals)
                 if fisher_pval < alpha:
                     is_root = False
--- a/causallearn/search/HiddenCausal/GIN/GIN.py
+++ b/causallearn/search/HiddenCausal/GIN/GIN.py
@@ -35,16 +35,25 @@ def GIN(data, indep_test_method='kci', alpha=0.05):
     -------
     G : general graph
         causal graph
-    K : list
+    causal_order : list
         causal order
     '''
     n = data.shape[1]
     cov = np.cov(data.T)
 
     if indep_test_method == 'kci':
-        indep_test = KCI_UInd()
-    else:
+        kci = KCI_UInd()
+
+    if indep_test_method not in ['kci', 'hsic']:
         raise NotImplementedError((f"Independent test method {indep_test_method} is not implemented."))
+
+    def indep_test(x, y, method=indep_test_method):
+        if method == 'kci':
+            return kci.compute_pvalue(x, y)[0]
+        elif method == 'hsic':
+            return hsic_test_gamma(x, y)[1]
+        else:
+            raise NotImplementedError((f"Independent test method {indep_test_method} is not implemented."))
 
     var_set = set(range(n))
     cluster_size = 2
@@ -56,7 +65,7 @@ def GIN(data, indep_test_method='kci', alpha=0.05):
             e = cal_e_with_gin(data, cov, list(cluster), list(remain_var_set))
             pvals = []
             for z in range(len(remain_var_set)):
-                pvals.append(indep_test.compute_pvalue(data[:, [z]], e[:, None])[0])
+                pvals.append(indep_test(data[:, [z]], e[:, None]))
             fisher_pval = fisher_test(pvals)
             if fisher_pval >= alpha:
                 tmp_clusters_list.append(cluster)
@@ -66,13 +75,13 @@ def GIN(data, indep_test_method='kci', alpha=0.05):
             var_set -= set(cluster)
         cluster_size += 1
 
-    K = []
+    causal_order = [] # this variable corresponds to K in paper
     updated = True
     while updated:
         updated = False
         X = []
         Z = []
-        for cluster_k in K:
+        for cluster_k in causal_order:
             cluster_k1, cluster_k2 = array_split(cluster_k, 2)
             X += cluster_k1
             Z += cluster_k2
@@ -87,13 +96,13 @@ def GIN(data, indep_test_method='kci', alpha=0.05):
                 e = cal_e_with_gin(data, cov, X + cluster_i1 + cluster_j1, Z + cluster_i2)
                 pvals = []
                 for z in range(len(Z + cluster_i2)):
-                    pvals.append(indep_test.compute_pvalue(data[:, [z]], e[:, None])[0])
+                    pvals.append(indep_test(data[:, [z]], e[:, None]))
                 fisher_pval = fisher_test(pvals)
                 if fisher_pval < alpha:
                     is_root = False
                     break
             if is_root:
-                K.append(cluster_i)
+                causal_order.append(cluster_i)
                 clusters_list.remove(cluster_i)
                 updated = True
                 break
@@ -106,7 +115,7 @@ def GIN(data, indep_test_method='kci', alpha=0.05):
     latent_id = 1
     l_nodes = []
 
-    for cluster in K:
+    for cluster in causal_order:
         l_node = GraphNode(f"L{latent_id}")
         l_node.set_node_type(NodeType.LATENT)
         G.add_node(l_node)
@@ -140,7 +149,7 @@ def GIN(data, indep_test_method='kci', alpha=0.05):
             G.add_directed_edge(l_node, o_node)
         latent_id += 1
 
-    return G, K
+    return G, causal_order
 
 
 def GIN_MI(data):
@@ -157,7 +166,7 @@ def GIN_MI(data):
     -------
     G : general graph
         causal graph
-    K : list
+    causal_order : list
         causal order
     '''
     v_labels = list(range(data.shape[1]))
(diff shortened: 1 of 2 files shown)


tests/TestGIN.py
```diff
@@ -1,104 +1,81 @@
...
+        L2 = np.random.uniform(1.2, 1.8) * L1 + np.random.uniform(-1, 1, size=sample_size)
+        X1 = np.random.uniform(1.2, 1.8) * L1 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
+        X2 = np.random.uniform(1.2, 1.8) * L1 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
+        X3 = np.random.uniform(1.2, 1.8) * L2 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
+        X4 = np.random.uniform(1.2, 1.8) * L2 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
         data = np.array([X1, X2, X3, X4]).T
         data = (data - np.mean(data, axis=0)) / np.std(data, axis=0)
-        g, k = GIN(data)
-        print(g, k)
+        _, k = GIN(data)
```
tofuwen, line 22: please give better naming (instead of naming it "k")

Basically you should never use not-meaningful variable name like "k" (only exception is well-known i and j for loops)

tests/TestGIN.py
```diff
@@ -1,104 +1,81 @@
...
+        X5 = np.random.uniform(1.2, 1.8) * L2 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
+        X6 = np.random.uniform(1.2, 1.8) * L2 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
+        X7 = np.random.uniform(1.2, 1.8) * L3 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
+        X8 = np.random.uniform(1.2, 1.8) * L3 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
+        X9 = np.random.uniform(1.2, 1.8) * L3 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
         data = np.array([X1, X2, X3, X4, X5, X6, X7, X8, X9]).T
         data = (data - np.mean(data, axis=0)) / np.std(data, axis=0)
-        g, k = GIN(data)
-        print(g, k)
+        _, k = GIN(data)
```
tofuwen, line 48: same

tests/TestGIN.py
```diff
@@ -1,104 +1,81 @@
...
+        k = [sorted(k_i) for k_i in k]
```
tofuwen, line 49: previously you use i, here you use k_i. 

It would be better if you give consistent naming

causallearn/search/HiddenCausal/GIN/GIN.py
```diff
@@ -50,6 +41,11 @@ def GIN(data, indep_test=kci, alpha=0.05):
     n = data.shape[1]
     cov = np.cov(data.T)
 
+    if indep_test_method == 'kci':
+        indep_test = KCI_UInd()
+    else:
+        raise NotImplementedError((f"Independent test method {indep_test_method} is not implemented."))
+
```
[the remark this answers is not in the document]
zhi-yi-huang: This requires kernel-based independence tests such as KCI, HSIC. But they are called differently, and I wonder if I need to write a function to eliminate their differences.

causallearn/search/HiddenCausal/GIN/GIN.py
```diff
@@ -220,7 +212,7 @@ def cal_e_with_gin(data, cov, X, Z):
     cov_m = cov[np.ix_(Z, X)]
     _, _, v = np.linalg.svd(cov_m)
     omega = v.T[:, -1]
-    return np.dot(omega, data[:, X].T)
+    return np.dot(data[:, X], omega.T)
```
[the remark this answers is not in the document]
zhi-yi-huang: Reduce the overhead of data matrix transposition.

tests/TestGIN.py
zhi-yi-huang: Because 'i' means index, so change the i to k_i, I forgot to unify it when I changed it before.

tests/TestGIN.py
zhi-yi-huang: Thanks for the suggestion, I will update it later. I see that k is often used in papers to indicate causal order, e.g. [LiNGAM](https://www.jmlr.org/papers/v7/shimizu06a.html), so I just used k.

tests/TestGIN.py
```diff
@@ -1,104 +1,81 @@
...
-
+        sample_size = 500
+        random.seed(42)
+        np.random.seed(42)
+        L1 = np.random.uniform(-1, 1, size=sample_size)
+        L2 = np.random.uniform(1.2, 1.8) * L1 + np.random.uniform(-1, 1, size=sample_size)
+        X1 = np.random.uniform(1.2, 1.8) * L1 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
+        X2 = np.random.uniform(1.2, 1.8) * L1 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
+        X3 = np.random.uniform(1.2, 1.8) * L2 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
+        X4 = np.random.uniform(1.2, 1.8) * L2 + 0.2 * np.random.uniform(-1, 1, size=sample_size)
```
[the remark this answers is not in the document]
zhi-yi-huang: The previous data generation parameters were set according to the paper, but generating according to the paper does not seem to be fully identifiable. So I changed the data generation parameters.

causallearn/search/HiddenCausal/GIN/GIN.py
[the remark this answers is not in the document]
tofuwen: so will the old code support HSIC? Basically my point is that, we should never regress (i.e. previously supported functions are not supported anymore).

if the old code base only supports KCI, then it's fine

causallearn/search/HiddenCausal/GIN/GIN.py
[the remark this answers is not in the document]
tofuwen: will this change the function return result? 

The current return value is the old value transpose, right?

tests/TestGIN.py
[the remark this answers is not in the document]
tofuwen: Why is the old data not fully identifiable? We can prove it not identifiable or just our algorithm fail to do so? Sorry causal n00b lol

tests/TestGIN.py
tofuwen: hmmm, makes sense haha. But it would still be good to name it clearly in our codebase (you can add comment to say this variable corresponds to k in paper)

tests/TestGIN.py
tofuwen: got it --- please make them consistent

causallearn/search/HiddenCausal/GIN/GIN.py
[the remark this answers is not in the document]
zhi-yi-huang: Sorry, omega seems to be a 1-dimensional array and transposing the matrix is an invalid operation. The output result is unchanged. np.dot operation on N-dimensional array with 1-dimensional array is in accordance with the algorithm logic. Please refer to [numpy.dot](https://numpy.org/doc/stable/reference/generated/numpy.dot.html)

tests/TestGIN.py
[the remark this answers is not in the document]
zhi-yi-huang: It should be an independence test error caused by the data.

zhi-yi-huang:
## Updates

- Added tests to test GIN algorithm using the hsic independence test.
- Updated docstring
- Removed default parameter of indep_test
- Fixed tests

## Test Plan

```python
python -m unittest tests.TestGIN # should pass
```

image

tests/TestGIN.py
```diff
@@ -1,104 +1,154 @@
...
+        data = np.array([X1, X2, X3, X4]).T
+        data = (data - np.mean(data, axis=0)) / np.std(data, axis=0)
+        _, causal_order = GIN(data, indep_test_method='kci')
+        causal_order = [sorted(cluster_i) for cluster_i in causal_order]
+        ground_truth = [[0, 1], [2, 3]]
+        assert len(causal_order) == len(ground_truth)
+        for i in range(len(causal_order)):
+            assert np.isclose(causal_order[i], ground_truth[i]).all()
```
tofuwen, line 27: you don't need to use np.isclose() right?

causal_order must equal to ground_truth, right?

tests/TestGIN.py
```diff
@@ -1,104 +1,154 @@
...
 
+    def test_case1_hsic(self):
```
tofuwen, line 29: this function looks almost exactly the same as last function. 

Please consider re-use the code, instead of copy and paste (basically you should never copy and paste code)

You can refer to https://github.com/cmu-phil/causal-learn/pull/59 to make your test clearer.

tests/TestGIN.py
```diff
@@ -1,104 +1,154 @@
...
-        X8 = np.random.uniform(0.5, 2.0) * L3 + np.random.uniform(-1, 1, size=sample_size) ** 5
-        X9 = np.random.uniform(0.5, 2.0) * L3 + np.random.uniform(-1, 1, size=sample_size) ** 5
+        _, causal_order = GIN(data, indep_test_method='hsic')
+        causal_order = [sorted(cluster_i) for cluster_i in causal_order]
+        ground_truth = [[0, 1], [2, 3]]
+        assert len(causal_order) == len(ground_truth)
+        for i in range(len(causal_order)):
+            assert np.isclose(causal_order[i], ground_truth[i]).all()
 
+    def test_case2_kci(self):
```
tofuwen, line 48: same for test case 2.

Basically you can test both kci and hsic in a single function, and the only difference seems to be GIN(data, indep_test_method) indep_test_method this parameter. 

Check https://github.com/cmu-phil/causal-learn/pull/59

tests/TestGIN.py
```diff
@@ -1,104 +1,154 @@
...
+        data = (data - np.mean(data, axis=0)) / np.std(data, axis=0)
+        _, causal_order = GIN(data, indep_test_method='hsic')
+        causal_order = [sorted(cluster_i) for cluster_i in causal_order]
+        ground_truth = [[0, 1, 2], [3, 4, 5], [6, 7, 8]]
+        assert len(causal_order) == len(ground_truth)
+        for i in range(len(causal_order)):
+            assert np.isclose(causal_order[i], ground_truth[i]).all()
+
+
+    def test_case3_kci(self):
```
tofuwen, line 99: same

tofuwen (commented):
Thanks for the great work!

Almost done --- we should do a small refactor to make code more concise. 
Remember the principle: never copy and paste and re-use code whenever possible. :)

tofuwen:
BTW, I really like the three tests you designed. I think that's great. :)

Also a small nit: when you made later changes, you can update your description and test plan instead of commenting a new one --- people will generally read the first one instead of scroll down to check all the conversions. This makes this PR clearer for other people. :)

tofuwen:
@zhi-yi-huang do you mind addressing the comments? After that, I think we can merge this PR --- very close now

tests/TestGIN.py
zhi-yi-huang: For the first one, I don't understand. For the second one, yes.

tests/TestGIN.py
zhi-yi-huang: Thank you for your suggestion, I will refer to modify the test code.

zhi-yi-huang:
Sorry, I missed the email from GitHub.

tests/TestGIN.py
```diff
@@ -1,104 +1,88 @@
...
+        for indep_test_method in TestGIN.indep_test_methods:
+            _, causal_order = GIN(data, indep_test_method=indep_test_method, alpha=alpha)
+            causal_order = [sorted(cluster_i) for cluster_i in causal_order]
+            TestGIN.validate_result(ground_truth, causal_order)
+
+    @staticmethod
+    def validate_result(ground_truth, estimated_result):
+        assert len(ground_truth) == len(estimated_result)
+        for i in range(len(estimated_result)):
+            assert np.isclose(estimated_result[i], ground_truth[i]).all()
```
tofuwen, line 88: why do we use "isclose()" instead of "==" here? 

With integer comparison, you should expect "==", right?

I think "isclose()" is meant to compare floating point

tofuwen (commented):
Thanks for the great work, it looks much better now!

Finally one small thing need to be addressed

tofuwen:
This is awesome! I think this PR is ready to be merged. :)

cc @kunwuz

PR #62, opened 2022-07-19
Refactor CITs in oop way

MarkDana:
### Updated files:

+ `cit.py`: Last time we rewrite all cit functions into one `CIT` class, with all methods in one class. This time we further separate each test method into a subclass inherited from a base class `CIT_Base`.

### How to use the new class(es):
+ Code logic is consistent with the current version ([ac3f4e7](https://github.com/cmu-phil/causal-learn/commit/ac3f4e71d81bbafe2ecccef36012a108b01360fa)).
+ For users, any algorithms can be run end-to-end in the same way as before , e.g., (see [here](https://github.com/cmu-phil/causal-learn/pull/46#issue-1292042971))

```python
from causallearn.search.ConstraintBased.PC import pc
from causallearn.utils.cit import fisherz

cg = pc(data, 0.05, fisherz)
```
+ For developers, to declare a cit instance is also the same as before (so no other files are changed):

```python

fisherz_obj = CIT(data, "fisherz") # construct a CIT instance with data and method name
pValue = fisherz_obj(X, Y, S) # a simple call is ok. no need to consider cache/corr_mat etc. by yourself.
```

though in code before, `CIT` is a class while now `CIT` is a function API that returns the respective class. So an alternative way of writing code above is:

```python
from causallearn.utils.cit import FisherZ
fisherz_obj = FisherZ(data)
pValue = fisherz_obj(X, Y, S)
```

+ [Issues on MVPC's inaccurate fisherz result](https://github.com/cmu-phil/causal-learn/blob/ac3f4e71d81bbafe2ecccef36012a108b01360fa/causallearn/utils/cit.py#L129) is solved. It's due to samplesize's change (my fault). Code logic is consistent as before.

+ Functions for cit's resume-from-break-point is added in the `CIT_Base` class. I will create a new pr for reference.

### Test plan:
Same as [#46 Rewrite CITests as a class && re-use covariance matrix for fisherz](https://github.com/cmu-phil/causal-learn/pull/46):
```python
python -m unittest TestPC    # should pass
python -m unittest TestFCI    # should pass
python -m unittest TestCDNOD    # should pass
python -m unittest TestMVPC    # should pass
python -m unittest TestMVPC_mv_fisherz_test.py    # should pass
python -m unittest TestMVPC_mv_fisherz.py    # should pass except for test_pc_with_mv_fisherz_MCAR_data_assertion, which cannot pass in the original code.
```

### DIFF
--- a/causallearn/utils/cit.py
+++ b/causallearn/utils/cit.py
@@ -8,6 +8,7 @@ from causallearn.utils.KCI.KCI import KCI_CInd, KCI_UInd
 from causallearn.utils.PCUtils import Helper
 
 CONST_BINCOUNT_UNIQUE_THRESHOLD = 1e5
+NO_SPECIFIED_PARAMETERS_MSG = "NO SPECIFIED PARAMETERS"
 fisherz = "fisherz"
 mv_fisherz = "mv_fisherz"
 mc_fisherz = "mc_fisherz"
@@ -55,7 +56,7 @@ class CIT_Base(object):
         self.SAVE_CACHE_CYCLE_SECONDS = 30
         self.last_time_cache_saved = time.time()
         self.pvalue_cache = {'data_hash': self.data_hash}
-        if not cache_path is None:
+        if cache_path is not None:
             assert cache_path.endswith('.json'), "Cache must be stored as .json file."
             if os.path.exists(cache_path):
                 with codecs.open(cache_path, 'r') as fin: self.pvalue_cache = json.load(fin)
@@ -90,8 +91,8 @@ class CIT_Base(object):
 
         Parameters
         ----------
-        X: int, or np.*int*
-        Y: int, or np.*int*
+        X: int, or np.*int*, or Iterable<int | np.*int*>
+        Y: int, or np.*int*, or Iterable<int | np.*int*>
         condition_set: Iterable<int | np.*int*>
 
         Returns
@@ -132,7 +133,7 @@ class CIT_Base(object):
 class FisherZ(CIT_Base):
     def __init__(self, data, **kwargs):
         super().__init__(data, **kwargs)
-        self.check_cache_method_consistent('fisherz', -1)   # -1: no parameters can be specified for fisherz
+        self.check_cache_method_consistent('fisherz', NO_SPECIFIED_PARAMETERS_MSG)
         self.assert_input_data_is_valid()
         self.correlation_matrix = np.corrcoef(data.T)
 
@@ -192,7 +193,7 @@ class Chisq_or_Gsq(CIT_Base):
             return np.unique(column, return_inverse=True)[1]
         assert method_name in ['chisq', 'gsq']
         super().__init__(np.apply_along_axis(_unique, 0, data).astype(np.int64), **kwargs)
-        self.check_cache_method_consistent(method_name, -1)   # -1: no parameters can be specified for chisq/gsq
+        self.check_cache_method_consistent(method_name, NO_SPECIFIED_PARAMETERS_MSG)
         self.assert_input_data_is_valid()
         self.cardinalities = np.max(self.data, axis=0) + 1
 
@@ -339,7 +340,7 @@ class Chisq_or_Gsq(CIT_Base):
 class MV_FisherZ(CIT_Base):
     def __init__(self, data, **kwargs):
         super().__init__(data, **kwargs)
-        self.check_cache_method_consistent('mv_fisherz', -1)   # -1: no parameters can be specified for mv_fisherz
+        self.check_cache_method_consistent('mv_fisherz', NO_SPECIFIED_PARAMETERS_MSG)
         self.assert_input_data_is_valid(allow_nan=True)
 
     def _get_index_no_mv_rows(self, mvdata):
@@ -386,7 +387,7 @@ class MC_FisherZ(CIT_Base):
     def __init__(self, data, **kwargs):
         # no cache for MC_FisherZ, since skel and prt_m is provided for each test.
         super().__init__(data, **kwargs)
-        self.check_cache_method_consistent('mc_fisherz', -1)   # -1: no parameters can be specified for mc_fisherz
+        self.check_cache_method_consistent('mc_fisherz', NO_SPECIFIED_PARAMETERS_MSG)
         self.assert_input_data_is_valid(allow_nan=True)
         self.mv_fisherz = MV_FisherZ(data, **kwargs)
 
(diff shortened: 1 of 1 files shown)


causallearn/utils/cit.py
```diff
@@ -15,78 +15,142 @@
 chisq = "chisq"
 gsq = "gsq"
 
-class CIT(object):
-    def __init__(self, data, method='fisherz', **kwargs):
+
+def CIT(data, method='fisherz', **kwargs):
```
tofuwen, line 19: hmmm, I don't like this kargs design, because users doesn't know what to input?

I think the current way is fine (for backward compatibility), and later, maybe we can change it to 
cit = FisherZ(data, args) for all caller?

causallearn/utils/cit.py
```diff
@@ -15,78 +15,142 @@
...
-        method: str, in ["fisherz", "mv_fisherz", "mc_fisherz", "kci", "chisq", "gsq"]
-        kwargs: placeholder for future arguments, or for KCI specific arguments now
+        data: data matrix, np.ndarray, in shape (n_samples, n_features)
+        cache_path: str, path to save cache .json file. default as None (no io to local file).
+        kwargs: for future extension.
         '''
+        assert isinstance(data, np.ndarray), "Input data must be a numpy array."
         self.data = data
-        self.data_hash = hash(str(data))
+        self.data_hash = hashlib.md5(str(data).encode('utf-8')).hexdigest()
```
tofuwen, line 52: When data is huge, this will be slow? 

And when path is None, we don't need to compute data_hash?

causallearn/utils/cit.py
```diff
@@ -15,78 +15,142 @@
...
+            1. convert to built-in types for json serialization
+            2. handle multi-dim unconditional variables (for kernel-based)
+            3. basic check for valid input (X, Y no overlap with condition_set)
+            4. generate unique and hashable cache key
+
+        Parameters
+        ----------
+        X: int, or np.*int*
+        Y: int, or np.*int*
+        condition_set: Iterable<int | np.*int*>
```
tofuwen, line 95: why not force it to be List<int>?

causallearn/utils/cit.py
```diff
@@ -15,78 +15,142 @@
...
-            def _unique(column):
-                return np.unique(column, return_inverse=True)[1]
-            self.data = np.apply_along_axis(_unique, 0, self.data).astype(np.int64)
-            self.data_hash = hash(str(self.data))
-            self.cardinalities = np.max(self.data, axis=0) + 1
+        self.cache_path = cache_path
+        self.SAVE_CACHE_CYCLE_SECONDS = 30
+        self.last_time_cache_saved = time.time()
+        self.pvalue_cache = {'data_hash': self.data_hash}
+        if not cache_path is None:
```
tofuwen, line 58:
```suggestion
        if cache_path is not None:
```

check https://stackoverflow.com/questions/2710940/python-if-x-is-not-none-or-if-not-x-is-none

causallearn/utils/cit.py
```diff
@@ -15,78 +15,142 @@
...
+            # ulist1, ulist2, clist: list of ints, sorted.
+            _strlst  = lambda lst: '.'.join(map(str, lst))
+            return f'{_strlst(ulist1)};{_strlst(ulist2)}|{_strlst(clist)}' if len(clist) > 0 else \
+                   f'{_strlst(ulist1)};{_strlst(ulist2)}'
+
+        # every time when cit is called, auto save to local cache.
+        self.save_to_local_cache()
+
+        METHODS_SUPPORTING_MULTIDIM_DATA = ["kci"]
+        if condition_set is None: condition_set = []
```
tofuwen, line 114: might be cleaner to force condition_set is never None? 

If the user don't want condition_set, just use []?

causallearn/utils/cit.py
```diff
@@ -15,78 +15,142 @@
...
+        def _stringize(ulist1, ulist2, clist):
+            # ulist1, ulist2, clist: list of ints, sorted.
+            _strlst  = lambda lst: '.'.join(map(str, lst))
+            return f'{_strlst(ulist1)};{_strlst(ulist2)}|{_strlst(clist)}' if len(clist) > 0 else \
+                   f'{_strlst(ulist1)};{_strlst(ulist2)}'
+
+        # every time when cit is called, auto save to local cache.
+        self.save_to_local_cache()
+
+        METHODS_SUPPORTING_MULTIDIM_DATA = ["kci"]
```
tofuwen, line 113: only KCI can support multi-dim data?

causallearn/utils/cit.py
```diff
@@ -15,78 +15,142 @@
...
+        '''
+        reformat the input X, Y and condition_set to
+            1. convert to built-in types for json serialization
+            2. handle multi-dim unconditional variables (for kernel-based)
+            3. basic check for valid input (X, Y no overlap with condition_set)
+            4. generate unique and hashable cache key
+
+        Parameters
+        ----------
+        X: int, or np.*int*
```
tofuwen, line 93: I think X can be List<int>?

causallearn/utils/cit.py
```diff
@@ -15,78 +15,142 @@
...
+        condition_set = sorted(set(map(int, condition_set)))
+
+        # usually, X and Y are 1-dimensional index (in constraint-based methods)
+        if self.method not in METHODS_SUPPORTING_MULTIDIM_DATA:
+            X, Y = (int(X), int(Y)) if (X < Y) else (int(Y), int(X))
+            assert X not in condition_set and Y not in condition_set, "X, Y cannot be in condition_set."
+            return [X], [Y], condition_set, _stringize([X], [Y], condition_set)
+
+        # also to support multi-dimensional unconditional X, Y (usually in kernel-based tests)
+        Xs = sorted(set(map(int, X))) if isinstance(X, Iterable) else [int(X)]  # sorted for comparison
```
tofuwen, line 125: why not force X to be List<int> type?

Personally I always prefer to make the variable typed --- it can remove lots of potential bug and make the code much cleaner.

causallearn/utils/cit.py
```diff
@@ -15,78 +15,142 @@
...
         p : the p-value of the test
-        """
-        var = list((X, Y) + condition_set)
+        '''
+        Xs, Ys, condition_set, cache_key = self.get_formatted_XYZ_and_cachekey(X, Y, condition_set)
+        if cache_key in self.pvalue_cache: return self.pvalue_cache[cache_key]
+        var = Xs + Ys + condition_set
         sub_corr_matrix = self.correlation_matrix[np.ix_(var, var)]
         try:
             inv = np.linalg.inv(sub_corr_matrix)
```
tofuwen, line 156: hmmm, curious, what if exception is thrown here? You didn't catch?

causallearn/utils/cit.py
```diff
@@ -15,78 +15,142 @@
...
+        Ys = sorted(set(map(int, Y))) if isinstance(Y, Iterable) else [int(Y)]
+        Xs, Ys = (Xs, Ys) if (Xs < Ys) else (Ys, Xs)
+        assert len(set(Xs).intersection(condition_set)) == 0 and \
+               len(set(Ys).intersection(condition_set)) == 0, "X, Y cannot be in condition_set."
+        return Xs, Ys, condition_set, _stringize(Xs, Ys, condition_set)
+
+class FisherZ(CIT_Base):
+    def __init__(self, data, **kwargs):
+        super().__init__(data, **kwargs)
+        self.check_cache_method_consistent('fisherz', -1)   # -1: no parameters can be specified for fisherz
```
tofuwen, line 135: hmm, -1 here looks ugly. maybe we can have better design, e.g. use None?

And in the code below, it seems you never write "parameters_hash" to json? Why the assertion not fail?

causallearn/utils/cit.py
```diff
@@ -96,112 +160,43 @@ def fisherz(self, X, Y, condition_set):
...
+        return p
 
-    def _chisq_or_gsq_test(self, dataSXY, cardSXY, G_sq=False):
+class Chisq_or_Gsq(CIT_Base):
+    def __init__(self, data, method_name, **kwargs):
+        def _unique(column):
+            return np.unique(column, return_inverse=True)[1]
+        assert method_name in ['chisq', 'gsq']
+        super().__init__(np.apply_along_axis(_unique, 0, data).astype(np.int64), **kwargs)
+        self.check_cache_method_consistent(method_name, -1)   # -1: no parameters can be specified for chisq/gsq
```
tofuwen, line 195: same

causallearn/utils/cit.py
```diff
@@ -481,9 +562,4 @@ def __call__(self, X, Y, condition_set=None, *args):
 #     result = [[]]
 #     for pool in lists:
 #         result = [x + [y] for x in result for y in pool]
-#     return result
-
-
-
-
-
+#     return result
```
tofuwen, line 565: remove this to make the code cleaner

causallearn/utils/cit.py
MarkDana: I see your point! And I am also confused about this...

Now **kwargs is placed here to support some user-defined parameters (just at the algorithm call level, e.g., pc, s.t. users don't need to edit codes inside pc). E.g.,
 - `pc(data, 0.05, kci, est_width='median')` if the user wants to use pc+kci with another kernel width (which is not supported in the old `func kci`, where all parameters are set by default.)
 - `pc(data, 0.05, kci, True, 0, -1, cache_path='/my/path/to/cache.json')` to save&load citest cache.

Now I use kwargs because the additional arguments that different methods (e.g., FisherZ, KCI) can take are different, and we have to use `func CIT` as the entrance for all callers (just for backward compatibility).

args won't work because users still don't know what to input (and even the order)? A perfect way would be to declare cit outside the algorithm call:

```python
kci_obj = KCI(data, kernelZ='Polynomial', est_width='median', cache_path='/my/path/to/cache.json')
pc(data, 0.05, kci_obj, True, 0, -1) # only take the algorithm-related parameters
```

But this is not backward compatible (for user input). So a compromise might be to still use `func CIT` as entrance, and just put more instructions (on what parameters are allowable) at CIT's comment and CIT, pc's documents?

causallearn/utils/cit.py
MarkDana: It will be fast (and used only once). `np.ndarray.__str__` only returns a preview:

```bash
In [41]: data = np.random.randn(10000, 100)

In [42]: str(data)
Out[42]: '[[-1.39552534  1.37974053 -1.1619043  ...  0.13616104 -0.12120668\n  -1.00001339]\n [-0.25197878 -2.00971912  0.63008704 ... -0.97997436 -1.21297862\n   1.42272323]\n [-1.22421999  0.90022162 -1.33748472 ...  1.32908047 -1.37618144\n  -0.28312766]\n ...\n [ 1.71461535  0.10882434  0.08604805 ...  1.34678215 -2.30936746\n   0.76045509]\n [ 0.55727436  0.2203048   0.41242777 ...  0.95881301  0.58538315\n   1.26002782]\n [-0.77753666  0.53018912  0.70592259 ...  0.14847539 -0.60861808\n  -0.36093896]]'

In [43]: len(str(data))
Out[43]: 491

In [44]: timeit hashlib.md5(str(data).encode('utf-8')).hexdigest()
125 µs ± 392 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

causallearn/utils/cit.py
MarkDana: Wow, thanks for this!!! Marked!

I usually get lost when writing something like this...

causallearn/utils/cit.py
MarkDana: For now, yes.

causallearn/utils/cit.py
MarkDana: In all of our constraint-based methods, X and Y are assumed to be integers.

If we force X as List type, all the codes related to cit calls in pc, fci, ... will need to be changed.

Overall, integers X and Y should always be the first-class citizen in CITests. Multi-dim X Y only works in KCI, and not in constaint-based methods, but somewhere else (e.g., GIN, you name it).

causallearn/utils/cit.py
MarkDana: Yeh -1 is ugly, lol. A message string ("NO SPECIFIED PARAMETERS") might also be ok.

"parameters_hash" is written to cache (and json). See [`check_cache_method_consistent`](https://github.com/cmu-phil/causal-learn/blob/3781daec48142c53b0243dc000d015fcc2498fb3/causallearn/utils/cit.py#L65).

causallearn/utils/cit.py
MarkDana: Oh yes I didn't catch exceptions. But what would be the possible exceptions here? If it's about the type of X, Y, condition_set, a built-in error message seems to be informative enough.

causallearn/utils/cit.py
tofuwen: ic, sounds good.

causallearn/utils/cit.py
tofuwen: why do you "try" then?

causallearn/utils/cit.py
tofuwen: yeah, a (const) string is much better.

try to avoid to use any magic number like -1 here in your code. :)

causallearn/utils/cit.py
MarkDana: ic. Cool!

causallearn/utils/cit.py
MarkDana: Oh I missed this.

Sometimes users might want to input `FisherZ(X, Y)` to test just unconditional independence.

From the users' end, I feel that this looks better than `FisherZ(X, Y, [])`, or `FisherZ(X, Y, ())` ... (which of course also works).

Another reason is that, I'm not sure whether there exist usages like `FisherZ(X, Y)` in current codes. lol

causallearn/utils/cit.py
MarkDana: Oh yes!! Thanks!!

causallearn/utils/cit.py
MarkDana: Oh you mean which lines? L155?

This is from the original code at [Fixed fisherz test](https://github.com/cmu-phil/causal-learn/commit/dd1a13edcadc3b50baa51053c6e91f702bbf06e3) (#58).

causallearn/utils/cit.py
tofuwen: @MarkDana how about add a todo here to remind us later we want to remove the kargs argument?

causallearn/utils/cit.py
tofuwen: ha, ic. I missed the later "except".... nvm

tofuwen (approved):
great, I think this PR is ready to be merged! cc @kunwuz

PR #64, opened 2022-07-26
Added a local cache resume example in TestPC

MarkDana:
### Updated files:

+ `TestPC.py`: added `test_pc_with_citest_local_checkpoint` as a usage example of save&load local cache checkpoints. Other usages on constraint-based methods (e.g., FCI and CDNOD) can be just referred to here.
+ `PC.py`, `FCI.py`, `CDNOD.py`: added kwargs to `func pc` to pass additional parameters to CIT (and possibly more for future usage).

### Why we need this:
+ A typical use case 1: `pc(uc_rule=0, ...)` then after the long run and viewing the results, I want to just finetune some params (e.g., `pc(uc_rule=2, ...)` - then I'll need to wait for another 6 hrs??? (though most of the CI tests (on skeleton discovery) are the same).
+ A typical use case 2: sometimes due to slow speed/some random error I interrupt the running PC. Then I need to run it again - I've already spent hours on it, and want to resume from the breaking point.
+ In the above cases, usually the CIT results can be resumed, so as to save almost all of the running time - especially when the graph scale is big or KCI is used - where CI tests consume almost all time.
+ So, we designed a user-specified feature (default as off), to write CITest cache from runtime (which we already have at #6) to some user-specified local path (which we implemented at #62).

### How to use this:
Just refer to `test_pc_with_citest_local_checkpoint` in `TestPC.py`:

For example, if we plan to use local cache, instead of directly use

```python
cg = pc(data, 0.05, kci)
```

as before, now we use

```python
citest_cache_file = "./TestData/citest_cache_linear_10_first_500_kci.json"    # .json file
cg = pc(data, 0.05, kci, cache_path=citest_cache_file)
```

If `citest_cache_file` does not exist in your local cache, a new one will be created. Otherwise, the cache will be first loaded from the json file to the CIT class and used during the runtime. Note that 1) data hash and parameters hash will first be checked at loading to ensure consistency, and 2) during runtime, the cache will be saved to the local file every 30 seconds.

### Test plan:

```python
python -m unittest TestPC.TestPC.test_pc_with_citest_local_checkpoint
```

The test should pass and console will show:

```
First pc run takes 125.663s.
Second pc run takes 27.316s.
```

Also a `"./TestData/citest_cache_linear_10_first_500_kci.json"` is saved. It will look like:

```json
{
  "data_hash": "41b2bb03b69f7aa5910437ce481cd09a",
  "method_name": "kci",
  "parameters_hash": "99914b932bd37a50b983c5e7c90ae93b",
[...]

PR #65, opened 2022-07-26
Use d-separation as CIT in tests, to ensure PC's correctness

MarkDana:
### Note first: ###
This pr is based off of #64 to prevent conflict on `TestPC.py`. So please review #64 first.

### Updated files: ###
+ `cit.py`: add `class D_Separation(CIT_Base)`.
+ `TestPC.py`: add `test_pc_load_bnlearn_graphs_with_d_separation`.

### Test plan: ###

```
python -m unittest TestPC.TestPC.test_pc_load_bnlearn_graphs_with_d_separation
```

It should pass (but in long time (around 3hrs), since in big graphs: 1) more tests are needed, and 2) d-separation check takes more time to traverse paths). Here is an example output:

```
Depth=3, working on node 7: 100%|██████████| 8/8 [00:00<00:00, 6425.59it/s]
asia (8 nodes/8 edges): used 0.03562s, SHD: 0
Depth=3, working on node 4: 100%|██████████| 5/5 [00:00<00:00, 5866.16it/s]
cancer (5 nodes/4 edges): used 0.00914s, SHD: 0
Depth=3, working on node 4: 100%|██████████| 5/5 [00:00<00:00, 6878.16it/s]
earthquake (5 nodes/4 edges): used 0.00925s, SHD: 0
Depth=6, working on node 10: 100%|██████████| 11/11 [00:00<00:00, 5876.62it/s]
sachs (11 nodes/17 edges): used 0.13420s, SHD: 0
Depth=3, working on node 5: 100%|██████████| 6/6 [00:00<00:00, 7246.13it/s]
survey (6 nodes/6 edges): used 0.01338s, SHD: 0
Depth=5, working on node 36: 100%|██████████| 37/37 [00:00<00:00, 8497.93it/s]
alarm (37 nodes/46 edges): used 7.56482s, SHD: 0
Depth=7, working on node 47: 100%|██████████| 48/48 [00:00<00:00, 3848.28it/s]
barley (48 nodes/84 edges): used 701.48529s, SHD: 0
Depth=7, working on node 19: 100%|██████████| 20/20 [00:00<00:00, 7347.47it/s]
child (20 nodes/25 edges): used 0.97726s, SHD: 0
Depth=8, working on node 26: 100%|██████████| 27/27 [00:00<00:00, 7440.13it/s]
insurance (27 nodes/52 edges): used 21.79384s, SHD: 0
Depth=7, working on node 31: 100%|██████████| 32/32 [00:00<00:00, 4703.78it/s]
water (32 nodes/66 edges): used 122.41981s, SHD: 0
Depth=16, working on node 55: 100%|██████████| 56/56 [00:00<00:00, 6383.16it/s]
hailfinder (56 nodes/66 edges): used 365.14324s, SHD: 0
Depth=18, working on node 69: 100%|██████████| 70/70 [00:00<00:00, 3397.89it/s]
hepar2 (70 nodes/123 edges): used 7126.62941s, SHD: 0
Depth=9, working on node 75: 100%|██████████| 76/76 [00:00<00:00, 7891.84it/s]
win95pts (76 nodes/112 edges): used 167.08235s, SHD: 0
test_pc_load_bnlearn_graphs_with_d_separation passed!

Ran 1 test in 8514.537s
```
### Good news from above: ###

By this integration test, we know that our current implementation of pc (or more specifically, at least `pc(stable=True, uc_rule=0, uc_priority=-1)`) is correct, i.e., it will return the true CPDAG with asymptotic CI tests.

### DIFF
--- a/causallearn/utils/cit.py
+++ b/causallearn/utils/cit.py
@@ -15,6 +15,7 @@ mc_fisherz = "mc_fisherz"
 kci = "kci"
 chisq = "chisq"
 gsq = "gsq"
+d_separation = "d_separation"
 
 
 def CIT(data, method='fisherz', **kwargs):
@@ -37,6 +38,8 @@ def CIT(data, method='fisherz', **kwargs):
         return MV_FisherZ(data, **kwargs)
     elif method == mc_fisherz:
         return MC_FisherZ(data, **kwargs)
+    elif method == d_separation:
+        return D_Separation(data, **kwargs)
     else:
         raise ValueError("Unknown method: {}".format(method))
 
@@ -450,3 +453,38 @@ class MC_FisherZ(CIT_Base):
 
         virtual_cit = MV_FisherZ(data_vir)
         return virtual_cit(0, 1, tuple(cond_set_bgn_0))
+
+
+class D_Separation(CIT_Base):
+    def __init__(self, data, true_dag=None, **kwargs):
+        '''
+        Use d-separation as CI test, to ensure the correctness of constraint-based methods. (only used for tests)
+        Parameters
+        ----------
+        data:   numpy.ndarray, just a placeholder, not used in D_Separation
+        true_dag:   nx.DiGraph object, the true DAG
+        '''
+        super().__init__(data, **kwargs)  # data is just a placeholder, not used in D_Separation
+        self.check_cache_method_consistent('d_separation', NO_SPECIFIED_PARAMETERS_MSG)
+        self.true_dag = true_dag
+        import networkx as nx; global nx
+        # import networkx here violates PEP8; but we want to prevent unnecessary import at the top (it's only used here)
+
+    def __call__(self, X, Y, condition_set=None):
+        Xs, Ys, condition_set, cache_key = self.get_formatted_XYZ_and_cachekey(X, Y, condition_set)
+        if cache_key in self.pvalue_cache: return self.pvalue_cache[cache_key]
+        p = float(nx.d_separated(self.true_dag, {Xs[0]}, {Ys[0]}, set(condition_set)))
+        # pvalue is bool here: 1 if is_d_separated and 0 otherwise. So heuristic comparison-based uc_rules will not work.
+
+        # here we use networkx's d_separation implementation.
+        # an alternative is to use causal-learn's own d_separation implementation in graph class:
+        #   self.true_dag.is_dseparated_from(
+        #       self.true_dag.nodes[Xs[0]], self.true_dag.nodes[Ys[0]], [self.true_dag.nodes[_] for _ in condition_set])
+        #   where self.true_dag is an instance of GeneralGrpah class.
+        # I have checked the two implementations: they are equivalent (when the graph is DAG),
+        # and generally causal-learn's implementation is faster.
+        # but just for now, I still use networkx's, for two reasons:
+        # 1. causal-learn's implementation sometimes stops working during run (haven't check detailed reasons)
+        # 2. GeneralGraph class will be hugely refactored in the near future.
+        self.pvalue_cache[cache_key] = p
+        return p
--- a/tests/TestPC.py
+++ b/tests/TestPC.py
@@ -5,7 +5,7 @@ import unittest
 import hashlib
 import numpy as np
 from causallearn.search.ConstraintBased.PC import pc
-from causallearn.utils.cit import chisq, fisherz, gsq, kci, mv_fisherz
+from causallearn.utils.cit import chisq, fisherz, gsq, kci, mv_fisherz, d_separation
 from causallearn.graph.SHD import SHD
 from causallearn.utils.DAG2CPDAG import dag2cpdag
 from causallearn.utils.TXT2GeneralGraph import txt2generalgraph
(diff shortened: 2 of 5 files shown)


causallearn/utils/cit.py
```diff
@@ -450,3 +453,38 @@ def __call__(self, X, Y, condition_set, skel, prt_m):
 
         virtual_cit = MV_FisherZ(data_vir)
         return virtual_cit(0, 1, tuple(cond_set_bgn_0))
+
+
+class D_Separation(CIT_Base):
```
tofuwen, line 458: this is really hacky... Normally we shouldn't write hacky code like this.

D_separation is not a Conditional Independence Test, right? We probably shouldn't inherit CIT_base class.

You can think about this OOP design --- normally we need to follow the logic, if D-Separation and CIT has something in common (like here, data is the same, and you call want to return some value), you can add another layer of abstraction under CIT_base. 

We need to strictly follow the logical structure in code whenever possible. :)

tests/TestPC.py
```diff
@@ -330,3 +330,55 @@ def test_pc_load_bnlearn_discrete_datasets(self):
             print(f'{bname} ({num_nodes_in_truth} nodes/{num_edges_in_truth} edges): used {cg.PC_elapsed:.5f}s, SHD: {shd.get_shd()}')
 
         print('test_pc_load_bnlearn_discrete_datasets passed!\n')
+
+    # Test the usage of local cache checkpoint (check speed).
+    def test_pc_with_citest_local_checkpoint(self):
```
tofuwen, line 335: I really like the tests here! 

Cheers! Now we can guarantee our PC algorithm is indeed written correctly. Great work!! :)

causallearn/utils/cit.py
MarkDana: Thanks so much. I see your point! (Sorry I just saw your message..

My current codes are mainly for convenience - so that we can call d-separation just as if we call a citest (same as fisherz or kci). D-separation indeed has many things in common with citest (e.g., i/o), though yes, logically it is not a citest.

By "another layer of abstraction under CIT_base", are you suggesting something like `CIT_base -> D_separation_base -> D_separation`? Then this looks almost the same as `CIT_base -> D_separation`?

tofuwen: OOP is not just chain, it should be a DAG.

For example, you can design things like:

Data_base -> CIT_base -> ....
Data_base -> D_separation

MarkDana: I see!

A bit confused: then maybe we'll have to move all of our functionalities in CIT_base (e.g., input check, cache, etc) to Data_base - while they are not attributes about data, and D_separation requires no data?

To me, here D_separation is more like a duck type? Though in definition, it is NOT a citest (not a statistical one but a graphical one), in our context (to test the algorithm's correctness), we call it, use it and evaluate it all like a citest.

tofuwen: I see. 

I really didn't think much --- yeah, you are right, D-separation requires no data.

My point is just: in OOP, just think about what needs to be abstracted and shared.

So what's common between D_separation and CIT_base? The cache, and other related utils. Then probably make a base named Cache_base, and maybe you can design things like:

Cache_base -> CIT_base -> ....
Cache_base -> D_separation

Usually don't inherit things that you don't need and don't make OOP design not consistent with the logical structure (hacky like this will usually create troubles in the future.).

MarkDana: Thx @tofuwen. Cool! `Cache_base -> CIT_base -> ....` and `Cache_base -> D_separation` now looks logically reasonable. Though practically I still have this concern:

What is the difference set `CIT_base\Cache_base`? In other words, what is something shared by `FisherZ` and `Chisq`, but not used in `D_separation`? There is only one thing, `data`.

Therefore, `Cache_base` should contain cache-related utilities and input/output checks, and `CIT_base\Cache_base` should be only about data. However, if we do so, some problems arise:

- Cache-related utilities and data are relatively coupled (e.g., data hash check, parameters check). It may not be clean or easy to decouple the two.
- `Cache_base` - or naming it more accurately, e.g., `Cache_for_constraint_base` - is it something deserving a base class treatment, or just some utility functions belonging to the CITs? It's natural to understand `KCI` as a child class of `CIT_base`, but it seems weird to see `CIT_base` as a child class of `Cache_base`. Even without cache, CIT is still CIT.
- After all, d-separation is here implemented only to check the algorithms' correctness in **tests**. It is not some key function for users. So, is it really necessary to do the above refactor (separate out a `Cache_base`) only for d-separation - while sacrificing all the other main-function parts (mentioned above in the 2nd point)?
- And, can d-separation be seen as a citest? Maybe it depends. On how general we see them. For example,
  > d-separation is a criterion for deciding, from a given a causal graph, whether a set X of variables is independent of another set Y, given a third set Z.
  (http://bayes.cs.ucla.edu/BOOK-2K/d-sep.html)

I will think more about how to put d-separation in our package in a both logically reasonable and functionally clean way.

tofuwen: Yeah, I agree with you for the most part. I think you convinced me: I agree that my suggestions seem to add lots of extra work to make the (not very necessary design) better, which I don't think justify the increasing complexity here.

tests/TestPC.py
```diff
@@ -330,3 +330,55 @@ def test_pc_load_bnlearn_discrete_datasets(self):
...
+            num_edges_in_truth = truth_dag.get_num_edges()
+            num_nodes_in_truth = truth_dag.get_num_nodes()
+
+            true_dag_netx = nx.DiGraph()
+            true_dag_netx.add_nodes_from(list(range(num_nodes_in_truth)))
+            true_dag_netx.add_edges_from(set(map(tuple, np.argwhere(truth_dag.graph.T > 0))))
+
+            data = np.zeros((100, len(truth_dag.nodes)))  # just a placeholder
+            cg = pc(data, 0.05, d_separation, True, 0, -1, true_dag=true_dag_netx)
+            shd = SHD(truth_cpdag, cg.G)
```
tofuwen, line 379: one final questions: why we don't have assert here?

Should we assert shd = 0 here?

tofuwen (commented):
Thanks for the great work! We are almost done!

tests/TestPC.py
MarkDana: Oh yes, we should! Thanks for pointing this out!

causallearn/utils/cit.py
MarkDana: Cool. Thanks so much for this! The separated class for d-separation that you suggested would still be the perfect one, as long as we had enough time - maybe in a future refactor on citest.

tofuwen (approved):
Thanks for the awesome work! cc @kunwuz to merge

PR #72, opened 2022-09-08
Adding auto test for PNL

ErdunGAO:
**Updates**

- Clean some useless codes
1. The MixGuaussian layer is replaced by using MLP.
2. Clean the `print` and `plot` functions for debugging.
3. Clean the normalization function.
- Correct errors.
1. Correct the error in calculating the Jacob.
- Add an abnormal points deletion function for data pre-processing.
- Add the test file for PNL.

**Test plan**

`python -m unittest tests.TestPNL # should pass`

image

**TODO**

- Optimize the codes to run faster.
- Add the test for real data (At present, the performance on real data is unsatisfactory.).


tofuwen:
Thanks @ErdunGAO for your awesome work!

Could you update your description to be more descriptive? (i.e. what bugs you fixed, what changes you made etc.)

And by looking at code --- it seems you changed lots of logic. Why? Is it because our previous implementation may have some errors?

And I still remembered you said the current version doesn't work for real data, is it still the case? Are we trying to fix this issue? :)

ErdunGAO:
> Thanks @ErdunGAO for your awesome work!
> 
> Could you update your description to be more descriptive? (i.e. what bugs you fixed, what changes you made etc.)
> 
> And by looking at code --- it seems you changed lots of logic. Why? Is it because our previous implementation may have some errors?
> 
> And I still remembered you said the current version doesn't work for real data, is it still the case? Are we trying to fix this issue? :)

Sorry for the late reply @tofuwen.

I will add more detailed descriptions for this new request. This will be finished in the coming one or two days.

Actually, I just cleared some useless codes, e.g., debugging output lines and the useless Gaussian mixture function, and added an outlier points deletion function. The model parameters were also updated to perform better.

For the real data case, our method works (distinguish the cause and effect) but the performance is not good. Yes, I am working on this issue.

causallearn/search/FCMBased/PNL/PNL.py
```diff
@@ -115,56 +100,54 @@ def nica_mnd(self, X, TotalEpoch, KofMix):
...
-            {'params': G3.parameters()}], lr=1e-4, betas=(0.9, 0.99))
+            {'params': G2.parameters()}], lr=1e-5, betas=(0.9, 0.99))
+
+        loss_all = torch.zeros(TotalEpoch)
+        loss_pdf_all = torch.zeros(TotalEpoch)
+        loss_jacob_all = torch.zeros(TotalEpoch)
 
         for epoch in range(TotalEpoch):
+            G1.zero_grad()
+            G2.zero_grad()
```
tofuwen, line 127: will this change the original logic?

causallearn/search/FCMBased/PNL/PNL.py
```diff
@@ -181,28 +164,28 @@ def cause_or_effect(self, data_x, data_y):
         pval_forward: p value in the x->y direction
         pval_backward: p value in the y->x direction
         '''
+        torch.manual_seed(0)
```
tofuwen, line 167: why we need to fix random seed in your algorithm?

tests/TestPNL.py
```diff
@@ -1,55 +1,55 @@
 import os
 import sys
 
-sys.path.append("")
+BASE_DIR = os.path.join(os.path.dirname(__file__), '..')
+sys.path.append(BASE_DIR)
```
tofuwen, line 5: don't do this --- check discussions in https://github.com/cmu-phil/causal-learn/pull/59

causallearn/search/FCMBased/PNL/PNL.py
ErdunGAO: In our method, we take the neural network to learn/approximate the nonlinear relationship.

causallearn/search/FCMBased/PNL/PNL.py
ErdunGAO: `zero_grad()` should be added to clear the gradients. This is an error of the previous implementation.

PR #85, opened 2022-11-04
Update PNL

ErdunGAO:
**Updates**
- Change the gradient descent training method to stochastic gradient descent (For acceleration).
- Delete the `dele_abnormal` function, which is useless since our new implementation is robust.
- Add the testing results of our method on the *real data* and an additional simulation case *simulation_dataset_3*.

**Description**
- Files in  data folder `TestData/`
  - Reduce the number of samples in the two previous simulation datasets from $1000$ to $200$, to make the test faster. (Actually, $200$ is enough for finding the causal direction of two variables.)
  - Add one more simulation dataset.
- `TestPNL.py` file
  - Increase the p-value threshold from $0.1$ to $0.5$.
  - Update the stored results since the previous datasets are changed.
  - Add the test functions for the third simulation dataset and the real data.
- `PNL.py` file
  - Add a `PairDataset(Dataset)` function (for batch training).
  - Reduce the total epochs from $100000$ to $3000$.
  - Codes in lines $96-107$ are cleaned. Actually, the code logic is not changed.
  - Change the learning rate from $1e-5$ to $1e-4$.
  - Delete the loss recording variables named loss_all, loss_pdf_all, and loss_jacob_all.
  - Variables $y1$, $y2$ are deleted. $y1$ is useless and $y2$ is actually the estimated noise. We use $e$ to replace $y2$ and $e_estimated$ to name the estimated noise.
  - `zero_grad()` of G1 and G2 is replaced by `optimizer.zero_grad()`.
  - The `dele_abnormal` function is deleted.

**Test plan**
`python -m unittest tests.TestPNL # should pass`

image


tests/TestPNL.py
```diff
@@ -6,24 +6,26 @@ class TestPNL(unittest.TestCase):
...
     # y_1 = (x + x**3 + e)**2
     # y_2 = np.exp(x**2 + e)
+    # y_3 = (np.exp(x) + e) ** 3
     # np.savetxt(r"TestData/pnl_simulation_1.txt", np.hstack([x, y_1]), delimiter=',')
     # np.savetxt(r"TestData/pnl_simulation_2.txt", np.hstack([x, y_2]), delimiter=',')
+    # np.savetxt(r"TestData/pnl_simulation_3.txt", np.hstack([x, y_3]), delimiter=',')
 
     # Set the threshold for independence test
-    p_value_threshold = 0.1 # useless now but left
+    p_value_threshold = 0.5 # useless now but left
```
tofuwen, line 21: this is the only concern I have.

Do we have enough evidence to claim something if we compare with threshold 0.5? This basically means you can have false positive error as large as 0.5?

tofuwen (approved):

tofuwen:
cc @kunwuz to merge

PR #92, opened 2023-01-28
Fix Typo for GES Algorithm

wean2016:
**Updates**
Fix typo in GES.py. In detail, we change the dict key, 'd_label', from data type 'str' to 'int'.
Add TestLocalScore.py for validating the correctness of the local score functions.

**Test plan**
python -m unittest tests.TestLocalScore # should pass
python -m unittest tests.TestGES # should pass
image
image


kunwuz:
Thanks! It looks perfect :)

PR #108, opened 2023-04-07
Fix a numerical issue in fisherz

MarkDana:
See issue https://github.com/py-why/causal-learn/issues/107 for details.

The numerical issue of |sample partial correlation|>=1 happens when the samplesize is very small, or the relationship is deterministic.

Abs set to 1. - machine epsilon.

No tests required.


kunwuz (approved):
Thanks so much for the quick solution!

PR #122, opened 2023-07-10
Small bug in GeneralGraph.is_child_of

JanMarcoRuizdeVargas:
I think i found a small bug, node1 is_child_of node2 should return true when the endpoint at node1 is and arrow, not a tail.


kunwuz:
Thanks so much for the fix! I believe it is also the case in #130 by @Boyle-Coffee . Please feel free to let me know if you think that PR has any issues. Thanks again for your effort :)

kunwuz:
Hi, I'm temporarily closing this since it seems to be resolved in another PR #130. Please feel free to let me know if you would like to reopen it. Thanks again for your contribution!

JanMarcoRuizdeVargas:
Sorry I meant to answer earlier but seems I forgot. I think pr #130 is a better fix than mine, so thats fine.

PR #125, opened 2023-07-31
feat: optionally disable progress bar in fci

tiagodsilva:
Hi! 

**Context.** I am iteratively applying the FCI algorithm to multiple data sets; at each iteration, a new progress bar is created and my stdout becomes increasingly polluted. 

**Problem.** Is there a workaround to disable such progress bars when using FCI? 

**Solution.** Include `show_progress` as an argument to `fci`; this boolean variable is inherited by the `fas` sub-routine and is used to decide whether to enable (True) or disable (False) the progress bar.  

By the way, great package!


kunwuz (approved):
Thanks! :)

PR #144, opened 2023-10-27
Fixed indexing error in dag2pag

PhilippFaller:
The indices in the `sepset` may not be compatible with the indices used in the `PAG`, when ` ruleR3(PAG, sepset, ...)` is called.
This is, because the indices in the `sepset` depend on the order of `nodes`.
On the other hand, the indices returned by the `PAG.node_map()`-call within `ruleR3` depend on the order of `observed_nodes`, which is random, as `nodes` are converted to a `set` before being passed to the constructor of the `PAG.`

This pull request proposes a simple reindexing before the call of `ruleR3` and `ruleR4b`.


zhi-yi-huang (approved):
Thanks so much!

kunwuz:
Thanks for your PR! And thanks so much @zhi-yi-huang for the prompt review :)

PR #161, opened 2024-01-22
Fixed Issue#129

zhi-yi-huang:
# Updates
- Fixed an issue on updating the separated set. 
- Added a test for updating the separated set.

# Description

The `sepset` had been initialized with an empty set. But it should be recorded in `cg.sepset` when `x` and `y` are d-separated, given by `sepset`. Otherwise, `sepset` (the empty set) will be recorded in `cg.sepset`.

# Test Plan
``` sh
python -m unittest tests.TestPC # should pass
python -m unittest tests.TestSkeletonDiscovery # should pass
```
image


kunwuz (approved):

kunwuz:
Thanks for the fix!

PR #179, opened 2024-04-08
Update FCI.py for fixing isNoncollider

svenpieper:
Editing isNoncollider() for FCI Algo

Fixing Bug:
If `sep_sets` does not contain certain node combination, it will led to a `KeyError`.

With this fix, `None` will automatically set by the `get()` func. Also `get_node_map()` will only called once to check `set_sets` for content, thus make it a bit more efficient.


kunwuz (approved):
Thanks for your fix!

PR #192, opened 2024-08-22
Fixed a bug in finding possible parent sets in FCI

kenneth-lee-ch:
I realized there's a bug in the function `is_possible_parent` that I created for the orientation rules in FCI.


kunwuz (approved):
Thanks for finding and fixing it!

PR #193, opened 2024-08-25
fix: fix issue#186

zhi-yi-huang:
## Description
- Fixed a bug in `LocalScoreFunction.py`.

## Test Plan

``` sh
python -m unittest tests.TestLocalScore # should pass
```
image


kunwuz (approved):
Thanks so much for your fix!

PR #194, opened 2024-08-30
Add orientation rules 567 for Augmented FCI

EvieQ01:
Add orientation rules 567 for Augmented FCI according to [1]. These rules are related to undirected edges, in order to deal with selection bias. (when there's no selection, they can be ignored)

[1] Jiji Zhang, "On the completeness of orientation rules for causal discovery in the presence of latent confounders and selection bias", 2008


causallearn/search/ConstraintBased/FCI.py
```diff
@@ -26,8 +39,17 @@ def traverseSemiDirected(node: Node, edge: Edge) -> Node | None:
             return edge.get_node1()
     return None
 
+def traverseCircle(node: Node, edge: Edge) -> Node | None:
+    if node == edge.get_node1():
+        if edge.get_endpoint1() == Endpoint.CIRCLE and edge.get_endpoint2() == Endpoint.CIRCLE:
+            return edge.get_node2()
+    elif node == edge.get_node2():
+        if edge.get_endpoint1() == Endpoint.CIRCLE or edge.get_endpoint2() == Endpoint.CIRCLE:
```
zhi-yi-huang, line 47: Should the if statement condition here be an AND condition?

causallearn/search/ConstraintBased/FCI.py
```diff
@@ -371,6 +428,126 @@ def ruleR3(graph: Graph, sep_sets: Dict[Tuple[int, int], Set[int]], bk: Backgrou
...
+                                          and graph.get_endpoint(node, node_B) == Endpoint.CIRCLE and graph.get_endpoint(node_B, node) == Endpoint.CIRCLE]
+
+                #  get the adjacent nodes with circle edges of A and B that is non adjacent to B and A, respectively
+                for node_C in a_circle_adj_nodes_set:
+                    if graph.is_adjacent_to(node_B, node_C):
+                        continue
+                    for node_D in b_circle_adj_nodes_set:
+                        if graph.is_adjacent_to(node_A, node_D):
+                            continue
+                        path = GetUncoveredCirclePath(node_from=node_C, node_to=node_D, G=graph)
```
zhi-yi-huang, line 470: Will there be multiple uncovered circle paths? Perhaps we can change return to yield in line 105 to make it a generator.

causallearn/search/ConstraintBased/FCI.py
```diff
@@ -371,6 +428,126 @@ def ruleR3(graph: Graph, sep_sets: Dict[Tuple[int, int], Set[int]], bk: Backgrou
...
+                        if path is not None:
```
zhi-yi-huang, line 471: The edge between alpha and beta is not modified inside the if statement. Perhaps the edge between alpha and beta can be modified based on change_flag outside the if statement.

EvieQ01: I think the edge between alpha and beta will only be modified "if there is an uncovered circle path p s.t. xxx", so then change_flag should still be kept inside the if statement?

causallearn/search/ConstraintBased/FCI.py
EvieQ01: Thank you for the suggestion. This part is modified accordingly.

causallearn/search/ConstraintBased/FCI.py
zhi-yi-huang: Yes, I mean the edge between alpha and beta hasn't been modified. If change_flag is set to True, then there is an uncovered circle path between alpha and beta, and the edge between alpha and beta needs to be modified. Perhaps a change_flag can be set outside the for loop on line 443 to indicate that rule5 modifies graph G. And a local_change_flag can be set inside the for loop on line 446 to indicate that there is an uncovered circle path between alpha and beta, and the edge between alpha and beta can be modified according to the local_change_flag inside the for loop.

EvieQ01: I have added the part to allow orienting multiple uncovered circle paths between alpha and beta, and moved the change_flag accordingly. Now it is changed only when double tails are oriented.

zhi-yi-huang (commented):
Thanks for adding the orientation rules for FCI. I have fixed a minor bug in rule 5 and regenerates the benchmark results for FCI. 

``` sh
python -m unittest tests.TestFCI # should pass
```

image

kunwuz (approved):

kunwuz:
Thanks a lot!! @EvieQ01 @zhi-yi-huang

Issue #185, opened 2024-07-10
use of deprecated `np.mat`


Alex-Markham:
Numpy 2.0 was released less than a month ago, resulting in some deprecated code in causal-learn. For example, see [Line 42 ](https://github.com/py-why/causal-learn/blob/ae2f5b2a955b3ab7f6a29746016bed9a8f49b68a/causallearn/utils/ScoreUtils.py#L42)of `causal-learn/utils/ScoreUtils.py` which uses the deprecated `np.mat`. This should be a quick fix, so I'll try to make a (backwards-compatible) pull request in the next week or so, unless someone else gets to it first. 

It could also be worth considering adding [compatible release version specifiers](https://peps.python.org/pep-0440/#version-specifiers) in `setup.py` to ensure automatically installed requirements are compatible with versions that devs have used in tests.

kunwuz:
Aha, thank you so much for the reminder and your proposal to address it. The recent PR for GRaSP and BOSS (#188) removed some 'np.mat' in the local score functions, but there are likely still instances elsewhere. Please feel free to submit a PR, and let us know if you have any questions.

Alex-Markham:
I went through and removed every instance of `np.mat`, specifying the dimensions of the arrays and updating the surrounding code as needed. All of the tests are passing now except for `test_local_score_general` in `TestLocalScore.py`. I've done what I can, and hope it helps, but I think it still needs someone else to carefully go through `LocalScoreFunction.py` to fix the last failing test and address some related runtime warnings.

kunwuz:
Many thanks for your PR!! :)

PR #195, opened 2024-09-09
fix: deprecated `np.mat`

Alex-Markham:
#185


zhi-yi-huang (commented):
Thank you so much for your fix! I fix the conflict and bugs in `LocalScoreFunction.py`.

image

kunwuz (approved):
Thanks so much @Alex-Markham and @zhi-yi-huang !

PR #200, opened 2024-10-17
fixed the np.prod overflow issue in Chisq test

MarkDana:
Fixed https://github.com/py-why/causal-learn/issues/197

By allowing `np.prod' to return negative overflowed numbers.


kunwuz (approved):

kunwuz:
Thanks!

PR #201, opened 2024-10-21
Add two variants of the KCI test

OliverSchacht:
I am adding two variants of the kernel-based conditional independence test, that are aiming at improved computational efficiency.

Namely:
- The [Randomized Conditional Independence Test (RCIT)](https://arxiv.org/pdf/1702.03877 ) translated from [this R code](https://github.com/ericstrobl/RCIT/tree/master).
- The Fast and Parallel Kernel-based Conditional Independence test (FastKCI), implemented based on an idea of @Biwei-Huang and myself.

I also added two easy unit tests checking the implementation.
I added the algorithms to the cit class, so they can be used with the PC algorithm for causal discovery.

Happy to hear your feedback!

Edit: I think I got something with the signature wrong here... followed the guide but I think something is wrong.


kunwuz:
Hi Oliver, thanks for these implementations—they're really interesting and helpful! If possible, could you create a new PR or modify this one to include only the files with changes? It seems there's an issue with the commit history. You can ignore the signature; I can handle that manually.

OliverSchacht:
Hi!

Thanks for the feedback. Yes, I think I messed everything up by trying to rebase and sign. Never worked with that before... 
I will make a clean branch with only my additions and create a new PR.

PR #202, opened 2024-10-22
Add two variants of the KCI test

OliverSchacht:
This PR is updating #201 so it should be easier to merge.

Original Text: 
I am adding two variants of the kernel-based conditional independence test, that are aiming at improved computational efficiency.

Namely:
- The [Randomized Conditional Independence Test (RCIT)](https://arxiv.org/pdf/1702.03877 ) translated from [this R code](https://github.com/ericstrobl/RCIT/tree/master).
- The Fast and Parallel Kernel-based Conditional Independence test (FastKCI), implemented based on an idea of @Biwei-Huang and myself.

I also added two easy unit tests checking the implementation.
I added the algorithms to the cit class, so they can be used with the PC algorithm for causal discovery.

Happy to hear your feedback!

### DIFF
--- a/causallearn/utils/FastKCI/FastKCI.py
+++ b/causallearn/utils/FastKCI/FastKCI.py
@@ -9,19 +9,21 @@ from sklearn.gaussian_process import GaussianProcessRegressor
 from sklearn.gaussian_process.kernels import ConstantKernel, WhiteKernel, RBF
 import warnings
 
+
 class FastKCI_CInd(object):
     """
-    Python implementation of as speed-up version of the Kernel-based Conditional Independence (KCI) test. Unconditional version.
+    Python implementation of as speed-up version of the Kernel-based Conditional Independence (KCI) test.
+    Unconditional version.
 
     References
     ----------
     [1] K. Zhang, J. Peters, D. Janzing, and B. Schölkopf,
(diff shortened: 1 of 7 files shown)


MarkDana:
Dear @OliverSchacht I really appreciate this contribution!

I reviewed the codes and the logic seems good to me. I didn't identify major running issues. The two unittests passed from my end, though I didn't do other accuracy benchmarks (e.g, those in `TestPC.py/test_pc_simulate_linear_nongaussian_with_kci`).

(One minor point is that I see `KCI.py` is changed on `null_samples` -> `self.null_samples`. This change seems unrelated. Though trivial, I revert it just for sanity check. Two unittests still passed after this reversion. But please correct me if i'm wrong.)

For now I think this pr is ready to be merged. One suggestion: could you please also add documents to https://github.com/py-why/causal-learn/tree/main/docs/source/independence_tests_index, with general instructions on using FastKCI and RCIT, with parameter choices on e.g., `K`, `J`. This could be done maybe after the working paper is finished.

Thank you again :)

MarkDana (approved):

kunwuz:
Hi @OliverSchacht , thanks so much for your contribution! When you have time, could you please write a short document as @MarkDana suggested? This would be very helpful for our users to start using this excellent test.

PR #206, opened 2024-11-13
Graph operations compatible with np array

sherryzyh:
* Initialize Dag with numpy array
* DAG2CPDAG supporting numpy array


kunwuz (approved):
Thanks so much for your contribution!

PR #211, opened 2025-01-03
fix: handle residuals array is empty

emoen:
What if the correlation matrix is singular? In ExactSearch.bic_exact_search() handle residuals array is empty. It means that the least squares problem does not have a unique solution or has no degrees of freedom left to compute residuals. Return BIC score of negative infinity.


kunwuz (approved):
Thanks for the fix!

PR #215, opened 2025-01-10
Update default score function in BOSS.py

kunwuz:
Change the default score function to local_score_BIC since local_score_BIC_from_cov requires a different type of input.


bja43:
Just a warning, this will make things much slower.

bja43 (approved):

PR #245, opened 2025-10-19
Relabel nodes in nx_graph with labels

pbdanny:


kunwuz:
Aha, thanks for the careful thought. Since later we use integers to add edges, perhaps applying relabeling before that can unintentionally mix integer and string node types, which may lead to potential inconsistency. The existing self.labels mappings already provide a way to reference node names if needed, so we may want to keep the graph structure unchanged here.

PR #263, opened 2026-05-20
Fix NumPy scalar extraction in BIC scores

### LINKED ISSUE #262: Bug report: LocalScoreFunction slightly broken.
There appears to be a bug in LocalScoreFunction, though it should be easily fixable. I'm not a causal-learn contributor so I'm not submitting a PR, but the fix appears to be easy. Here's a minimal script that highlights the bug if you run it This will have affected any method in causal-learn that uses a LG score, possibly elsewhere as well, I don't know.

```
"""
Minimal BOSS bug reproducer for causal-learn.
No py-tetrad / JPype dependency.
"""

import numpy as np
from causallearn.search.PermutationBased.BOSS import boss

# Small random LG-SEM data — just enough to trigger the bug
rng = np.random.default_rng(42)
n_samples, n_nodes = 200, 10
data = rng.standard_normal((n_samples, n_nodes))

print("Python version:", __import__("sys").version)
print("NumPy version: ", np.__version__)
# print("causal-learn version:", __import__("causallearn").__version__)
print()
print("Running BOSS...")

cg = boss(data, score_func="local_score_BIC")

print("BOSS completed without error.")
print(f"Edges in result: {len(cg.G.get_graph_edges())}")

```

The fix should be simple; the problem occurs with NumPy ≥ 2.0, which has been the case for all recent releases. The identical offending line appears in both local_score_BIC and local_score_BIC_from_cov:

```
# Before (both functions):

sigma = float(cov[i, i] - yX @ XX_inv @ yX.T)

# After (both functions):

sigma = float((cov[i, i] - yX @ XX_inv @ yX.T).item())
```

The problem is, the expression yX @ XX_inv @ yX.T produces a 1×1 NumPy array, and subtracting it from the scalar cov[i, i] gives another 1×1 array. Python 3.14 / recent NumPy tightened float() to reject anything that isn't truly 0-dimensional, so .item() is needed to extract the scalar first before passing to float(). The rest of both functions is fine since np.log(sigma) and the arithmetic after it all handle plain Python floats without issue.

kunwuz:
Fix #262

### DIFF
--- a/causallearn/score/LocalScoreFunction.py
+++ b/causallearn/score/LocalScoreFunction.py
@@ -84,7 +84,7 @@ def local_score_BIC(Data: ndarray, i: int, PAi: List[int], parameters=None) -> f
             XX_inv = np.linalg.inv(XX)
         except np.linalg.LinAlgError:
             XX_inv = np.linalg.pinv(XX)
-        sigma = float(cov[i, i] - yX @ XX_inv @ yX.T)
+        sigma = float((cov[i, i] - yX @ XX_inv @ yX.T).item())
 
     if sigma <= 0:
         sigma = np.finfo(float).eps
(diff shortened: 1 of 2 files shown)


jdramsey (approved):
It looks good to me. My original minimal bug report script ran with a couple of adjustments for the recent repo. Of course it had zero edges in the model, so it was good that it got zero edges. I made another example with more edges, and it got a good answer. In any case, the problem preventing BOSS from running at all is fixed.

PR #267, opened 2026-06-17
Fix local_score_BDeu KeyError on pandas >= 2.0

v-shaal:
local_score_BDeu groups by a single parent column and calls GroupBy.get_group((ij,)). On pandas >= 2.0 a single grouper expects the scalar key, so the 1-tuple raises KeyError — making GES with score_func='local_score_BDeu' unusable on any current pandas (the bug is in the latest release 0.1.4.7 and on main).

Try the scalar key first and fall back to the 1-tuple form, so BDeu works on both old and new pandas. The resulting scores match a direct boolean-mask computation, and tests/TestGES.py::test_ges_load_discrete_10_with_local_score_BDeu runs to completion.


kunwuz (approved):
That's great, thanks!

